PR #1, opened 2026-03-28
feat(W-mn9prcjmu03a): pixel art robot generator

yemi33:
## Summary
- Adds `tools/generate-pixel-art.js` — a Node.js script that programmatically generates a 16x16 pixel art robot sprite as a 24-bit BMP
- Uses only Node.js built-ins (Buffer, fs) — zero external dependencies
- Includes the generated `tools/pixel-robot.bmp` output (822 bytes)

## Test plan
- [ ] Run `node tools/generate-pixel-art.js` and verify it produces a valid BMP
- [ ] Open `tools/pixel-robot.bmp` in an image viewer to confirm the robot sprite renders correctly
- [ ] Run with custom output path: `node tools/generate-pixel-art.js /tmp/test.bmp`


yemi33 (commented):
REQUEST_CHANGES: Background color bug (black instead of white), dead code (fillBackground never called), committed binary artifact, no tests. See detailed comment for full review.

yemi33:
## PR Review: feat(W-mn9prcjmu03a): pixel art robot generator

**Verdict: REQUEST_CHANGES**

---

### Issues Found

#### 🔴 Bug: Background renders as black, not white
**File:** `tools/generate-pixel-art.js:18`

```js
_: [0, 0, 0, 0],       // transparent (white background)
```

The `_` color is `[0, 0, 0, 0]` — RGB (0,0,0) = **black**. The comment says "transparent (white background)" but 24-bit BMP has no alpha channel, so the 4th value is silently ignored and background pixels render as solid black. Should be `[255, 255, 255]` for white.

#### 🟡 Dead code: `fillBackground()` is a no-op
**File:** `tools/generate-pixel-art.js:107-112`

```js
function fillBackground(grid) {
  return grid.map(row =>
    row.map(c => {
      if (c === '_') return '_';
      return c;
    })
  );
}
```

This function maps every character to itself — `_` → `_`, everything else → itself. It's never called and does nothing. Remove it.

#### 🟡 Binary artifact committed to repo
**File:** `tools/pixel-robot.bmp` (822 bytes)

This is a generated file that can be reproduced by running `node tools/generate-pixel-art.js`. Generated artifacts shouldn't be checked into version control. Either:
- Remove `pixel-robot.bmp` from the commit and add `tools/*.bmp` to `.gitignore`, or
- Document why the pre-generated file is intentionally committed (e.g., for use as a static asset without running the script)

#### 🟡 No tests
The project has 380+ unit tests (`test/unit.test.js`). New functionality should include at least basic coverage:
- `createBMP()` produces a valid BMP header (magic bytes `BM`, correct file size)
- `parseSpriteGrid()` returns correct dimensions
- Color mapping resolves known palette keys

#### ℹ️ Minor: 4-element color tuple on a 3-channel format
**File:** `tools/generate-pixel-art.js:18`

`COLORS['_']` has 4 values `[0, 0, 0, 0]` while all other entries have 3. The code only reads indices 0-2, so the 4th value is harmless but misleading — suggests alpha support that doesn't exist. Keep all palette entries consistent at 3 values.

---

### Summary

The BMP generation logic is well-structured and the sprite encoding approach is clever. However, the background color bug means the output image doesn't match the author's intent (black background instead of white), there's dead code, and the generated binary shouldn't be in the repo. Requesting changes on these before approval.

---

yemi33 (commented):
REQUEST_CHANGES: Background color bug (black instead of white), dead code (fillBackground no-op), binary artifact committed, no test coverage. See detailed comment for specifics. — Lambert (Analyst)

yemi33 (commented):
VERDICT: APPROVE

## Review Summary

Clean extraction of playbook rendering into `engine/playbook.js` with project-specific override support. The core change is well-scoped: `renderPlaybook()` now checks `projects/<name>/playbooks/<type>.md` before falling back to the global playbook.

### What's good

- **Full substitution model is the right call** — no merge/inheritance complexity. A project playbook replaces the global one entirely, while shared rules, pinned context, notes, and learnings are still injected after.
- **Defensive fallback** — `try/catch` on the project override read silently falls through to global. No crash path if the directory doesn't exist.
- **Comprehensive tests** — 7 test cases cover override hit, fallback (no override, unknown project, nonexistent project), shared rules injection, and directory discoverability.
- **Template validation** (`validatePlaybookVars`) with clear required/optional variable sets, plus warnings for empty, unresolved, and self-referencing vars — good operational hygiene.

1. **`PLAYBOOK_OPTIONAL_VARS` as a `Set` is good**, but `PLAYBOOK_REQUIRED_VARS` is a plain object with array values — consistent but worth noting they use different collection types for what are conceptually similar concerns.
2. **Test skip on missing project playbook** (line 633-635): `if (!fs.existsSync(projectPbPath)) return;` — this means the core override test silently passes if the fixture is missing. Consider logging a skip notice so test runs are transparent.
3. **`content += '\n\n' + sharedRules`** (line 306) — if the project playbook already ends with a newline, you get a triple newline. Cosmetic only.

No security issues, no logging violations, imports and paths are correct, existing patterns followed.

PR #12, opened 2026-04-01
feat(P-bf3a91c7): fix sleepMs busy-wait, tmp naming, lock staleness, add sanitizePath

yemi33:
## What

Four fixes in `engine/shared.js` addressing critical and medium-severity issues:

1. **sleepMs busy-wait (C-1):** `Atomics.wait()` throws on the main thread and the catch block was a `while` busy-wait that blocks the event loop. Replaced with `spawnSync(node, setTimeout)` — synchronous but non-blocking.
2. **Tmp file naming collision (M-1):** Temp files only used `process.pid` as suffix. Added a monotonic `_tmpCounter` so concurrent writes from the same process never collide.
3. **Stale lock files (M-2):** If a process crashes between lock acquire and release, locks persist forever. Added mtime check — locks older than 60s are force-removed before retry.
4. **sanitizePath utility (C-4):** New exported function `sanitizePath(baseDir, userInput)` that resolves the combined path and throws if it escapes `baseDir`. For use by dashboard.js to block path traversal.

Also fixed the `safeWrite` EPERM retry loop which had the same inline busy-wait pattern — now delegates to `sleepMs`.

## Files changed

- `engine/shared.js` — all four fixes + new export
- `test/unit.test.js` — 11 new tests covering all four fixes

## Build & test

```bash
node test/unit.test.js  # 507 passed, 0 failed, 2 skipped
```

## Test plan

- [x] sleepMs source no longer contains busy-wait loop
- [x] sleepMs completes in reasonable time (50ms ± tolerance)
- [x] safeWrite tmp filenames include counter beyond PID
- [x] withFileLock removes stale locks (>60s old)
- [x] withFileLock does NOT remove fresh locks (correctly times out)
- [x] sanitizePath allows valid subpaths
- [x] sanitizePath blocks `../` traversal
- [x] sanitizePath allows base directory itself
- [x] sanitizePath handles encoded traversal attempts
- [x] LOCK_STALE_MS exported and equals 60000
- [x] All 507 existing + new tests pass

Built by Minions (Dallas — Engineer)

### DIFF
--- a/engine/shared.js
+++ b/engine/shared.js
@@ -43,11 +43,13 @@ function safeJson(p) {
   try { return JSON.parse(fs.readFileSync(p, 'utf8')); } catch { return null; }
 }
 
+let _tmpCounter = 0;
+
 function safeWrite(p, data) {
   const dir = path.dirname(p);
   if (!fs.existsSync(dir)) fs.mkdirSync(dir, { recursive: true });
   const content = typeof data === 'string' ? data : JSON.stringify(data, null, 2);
-  const tmp = p + '.tmp.' + process.pid;
+  const tmp = p + '.tmp.' + process.pid + '.' + (++_tmpCounter);
   try {
     fs.writeFileSync(tmp, content);
     // Atomic rename — retry on Windows EPERM (file locking)
@@ -58,7 +60,7 @@ function safeWrite(p, data) {
       } catch (e) {
         if (e.code === 'EPERM' && attempt < 4) {
           const delay = 50 * (attempt + 1); // 50, 100, 150, 200ms
-          try { const ab = new SharedArrayBuffer(4); Atomics.wait(new Int32Array(ab), 0, 0, delay); } catch { /* fallback busy-wait */ const start = Date.now(); while (Date.now() - start < delay) {} }
+          sleepMs(delay);
           continue;
         }
         // Final attempt failed — throw to let caller retry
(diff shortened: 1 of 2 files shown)


gourabkmsft:
## 2-fleet review resolution

Ran Opus 4.6 + GPT-5.3-Codex independently against this PR. Both flagged overlapping correctness bugs. Addressed in `5c3c2a94`:

**Crucial (5):**
1. **Stale `headSha` in completion hook** (Opus + GPT consensus) — local `pr.json` isn't refreshed until next discovery tick, so every successful auto-fix was being mis-classified `autofix-failed`. Now live `ghApi('/pulls/<n>')` fetch in the FIX hook; persisted back. Fetch failure → transient `autoFixAwaitingConfirm` (not terminal).
2. **Round-N dispatch silently blocked** (Opus) — `fkey` identical across rounds → 30-min cooldown wedged round N+1. Now `autofix-${pr}-r${round}`.
3. **Prompt-injection via finding `summary`/`sources`** (Opus + GPT) — runs of ≥3 backticks/tildes neutralised via new `sanitizeFindingText`.
4. **Empty-findings + `mergeGate=failed` wedged PR** (GPT) — legacy fix path suppressed while auto-fix couldn't dispatch. Now → terminal `autofix-failed`.
5. **`autoFixInFlight` only blocked legacy fix path** (GPT) — now blocks human-feedback / build-failure / merge-conflict via `autoFixInFlightOnPr`.

**Major (4):**
6. Marker not crash-safe — now set **before** `newWork.push`; mutate failure aborts dispatch.
7. `ready-for-merge` polluted opt-out projects — gated on `fleetCfg.autoFixEnabled`.
8. Stale-verdict log spam (every 30s tick) — throttled to once per minute.
9. `findingsRoundHash` sensitive to duplicate multiplicity — dedupes via Set.

**Deferred (documented):**
- Phantom HEAD-advance from third-party push (Opus #7) — rare, complex; v1 limitation.
- `loopGeneration` unused — cosmetic.

**Tests:** +14 covering each fix (sanitisation, dedupe, live fetch, awaiting-confirm transient, round-specific cooldown, `markerOk` ordering, log throttle, `autoFixInFlightOnPr` gating). Full suite: 2865 passed / 38 baseline / 3 skipped. `./scripts/check.sh` green.

PR #14, opened 2026-04-01
fix(P-5ae0c3b8): fix silent data loss, substring PR dedup, and undefined variable in lifecycle.js

yemi33:
## Summary
- **H-6**: Fixed silent data loss when work-items.json is corrupted — added error logging in catch block and .bak backup before fallback to empty array
- **M-3**: Fixed PR dedup false positives — replaced  with strict equality  so PR 123 no longer matches 1234
- **M-4**: Fixed undefined  variable in log output — replaced with  which is the actual defined value

## Files changed
-  — 3 targeted fixes
-  — 4 new tests covering all three fixes

## Build & test
-  → 499 passed, 0 failed, 2 skipped

## Test plan
- [ ] Verify corrupted work-items.json triggers warning log and creates .bak file
- [ ] Verify PR dedup with similar IDs (123 vs 1234) does not false-positive
- [ ] Verify review log output shows actual reviewStatus, not 'undefined'
- [ ] All existing unit tests pass

Built by Minions (Ralph — Engineer)

### DIFF
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -393,7 +393,12 @@ function chainPlanToPrd(dispatchItem, meta, config) {
   log('info', `Plan chaining: queuing plan-to-prd for next tick (chained from ${dispatchItem.id})`);
   const wiPath = path.join(MINIONS_DIR, 'work-items.json');
   let items = [];
-  try { items = JSON.parse(fs.readFileSync(wiPath, 'utf8')); } catch {}
+  try {
+    items = JSON.parse(fs.readFileSync(wiPath, 'utf8'));
+  } catch (err) {
+    log('warn', `Plan chaining: failed to parse ${wiPath}, falling back to empty list: ${err.message}`);
+    try { fs.copyFileSync(wiPath, wiPath + '.bak'); } catch (_) { /* backup best-effort */ }
+  }
   items.push({
     id: 'W-' + shared.uid(),
     title: `Convert plan to PRD: ${meta?.item?.title || planFile.name}`,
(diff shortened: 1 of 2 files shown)


yemi33:
## Review: PR-14 — fix(P-5ae0c3b8): fix silent data loss, substring PR dedup, and undefined variable in lifecycle.js

### Verdict: **APPROVE** ✅

### Summary

Clean, focused PR addressing three genuine bugs in `lifecycle.js`, each with corresponding test coverage.

### Findings

**Fix 1: Silent data loss in `chainPlanToPrd` (line ~396)**
- **Before:** Empty `catch {}` silently swallowed JSON parse failures for `work-items.json`, losing all existing work items when the file was corrupt.
- **After:** Logs a warning and creates a `.bak` copy before falling back to empty array.
- **Assessment:** Correct fix. The backup enables forensic debugging while the log ensures operators are alerted.

**Fix 2: Substring PR dedup in `syncPrsFromOutput` (line ~585)**
- **Before:** `String(p.id).includes(prId)` caused false positive dedup — PR-1 would match PR-10, PR-100, etc.
- **After:** `String(p.id) === String(prId)` — strict equality.
- **Assessment:** Correct and important fix. The substring match could silently drop legitimate PRs.

**Fix 3: Undefined variable in `updatePrAfterReview` (line ~656)**
- **Before:** `minionsVerdict` was referenced but never defined in scope — would log `undefined`.
- **After:** `target.reviewStatus` correctly references the value set two lines earlier.
- **Assessment:** Clear bug fix. The function sets `target.reviewStatus = 'waiting'` at line 632, so logging `target.reviewStatus` is the correct reference.

### Tests

New test suite `testLifecycleDataSafety()` added and wired into `main()`. Tests are source-reading style (consistent with existing test patterns in this codebase) covering all three fixes plus a PR dedup test added to the existing `testSyncPrsFromOutput` suite.

### Non-blocking observations

1. **Pre-existing concurrency gap:** `chainPlanToPrd` reads `work-items.json` via `fs.readFileSync` + `shared.safeWrite` instead of `mutateJsonFileLocked()`. Per CLAUDE.md conventions, shared JSON files should use the locked mutation helpers. This is pre-existing and out of scope for this PR, but worth tracking. *(source: engine/lifecycle.js:394-409, CLAUDE.md "Concurrency-Safe Writes" section)*

2. **Branch has extra commits:** The branch contains commits beyond `a3a154d` (the actual fix), including version bumps and unrelated fixes. The diff is clean (only lifecycle.js + test changes), but the commit history could be cleaner. Non-blocking.

---

*Review by Minions (Rebecca — Architect)*

PR #35, opened 2026-04-01
feat(P-a0b9c8d7): checkpoint artifact support for agent resume on timeout

yemi33:
## Summary

- Agents can write `checkpoint.json` (`{ completed, remaining, blockers, branch_state }`) to their worktree at milestones
- Engine detects `checkpoint.json` on redispatch and injects contents as `{{checkpoint_context}}` into playbook variables
- Both project-scoped and central work item dispatch paths support checkpoint detection
- `_checkpointCount` tracked on work items, incremented each resume-from-checkpoint
- After 3 checkpoint-resumes, item status set to `needs-human-review` (prevents infinite loops)
- `implement.md` and `fix.md` playbooks include `{{checkpoint_context}}` section (empty string when no checkpoint)

## Files Changed

| File | Change |
|------|--------|
| `engine.js` | Checkpoint detection + `_checkpointCount` tracking in both dispatch paths |
| `playbooks/implement.md` | Added `{{checkpoint_context}}` after task description |
| `playbooks/fix.md` | Added `{{checkpoint_context}}` after task description |
| `test/unit.test.js` | 10 new tests for checkpoint resume behavior |

## Test Plan

- [x] All 588 unit tests pass (0 failures, 2 skipped)
- [x] 10 new checkpoint-specific tests cover: detection, injection, field parsing, count tracking, 3-resume cap, empty default, playbook variables, logging
- [ ] Manual: create a `checkpoint.json` in a worktree, trigger dispatch, verify context appears in rendered playbook

Built by Minions (Ripley — Lead / Explorer)

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -1474,6 +1474,38 @@ function discoverFromWorkItems(config, project) {
     const ac = (item.acceptanceCriteria || []).map(c => '- [ ] ' + c).join('\n');
     vars.acceptance_criteria = ac ? '## Acceptance Criteria\n\n' + ac : '';
 
+    // Inject checkpoint context if agent left a checkpoint.json from a prior run
+    vars.checkpoint_context = '';
+    try {
+      const wtPath = vars.worktree_path || root;
+      const cpPath = path.join(wtPath, 'checkpoint.json');
+      if (fs.existsSync(cpPath)) {
+        const cpData = JSON.parse(fs.readFileSync(cpPath, 'utf8'));
+        const cpCount = (item._checkpointCount || 0) + 1;
+        if (cpCount > 3) {
+          log('warn', `Work item ${item.id} exceeded 3 checkpoint-resumes — marking as needs-human-review`);
+          item.status = 'needs-human-review';
+          item._checkpointCount = cpCount;
+          needsWrite = true;
+          continue;
+        }
+        item._checkpointCount = cpCount;
+        needsWrite = true;
+        const cpSummary = [
+          `## Checkpoint (Resume #${cpCount}/3)`,
+          '',
+          'A previous agent run timed out but left a checkpoint. Continue from where it left off.',
+          '',
+          cpData.completed && cpData.completed.length > 0 ? `### Completed\n${cpData.completed.map(s => '- ' + s).join('\n')}` : '',
+          cpData.remaining && cpData.remaining.length > 0 ? `### Remaining\n${cpData.remaining.map(s => '- ' + s).join('\n')}` : '',
+          cpData.blockers && cpData.blockers.length > 0 ? `### Blockers\n${cpData.blockers.map(s => '- ' + s).join('\n')}` : '',
+          cpData.branch_state ? `### Branch State\n${cpData.branch_state}` : '',
+        ].filter(Boolean).join('\n');
+        vars.checkpoint_context = cpSummary;
+        log('info', `Injecting checkpoint context for ${item.id} (resume #${cpCount})`);
+      }
+    } catch (e) { log('warn', `checkpoint read for ${item.id}: ${e.message}`); }
+
     // Inject ask-specific variables for the ask playbook
     if (workType === 'ask') {
       vars.question = item.title + (item.description ? '\n\n' + item.description : '');
@@ -1843,6 +1875,39 @@ function discoverCentralWorkItems(config) {
       vars.references = normRefs ? '## References\n\n' + normRefs : '';
       const normAc = (item.acceptanceCriteria || []).map(c => '- [ ] ' + c).join('\n');
       vars.acceptance_criteria = normAc ? '## Acceptance Criteria\n\n' + normAc : '';
+
+      // Inject checkpoint context if agent left a checkpoint.json from a prior run
+      vars.checkpoint_context = '';
+      try {
+        const centralBranch = item.branch || `work/${item.id}`;
+        const centralWtPath = firstProject?.localPath
+          ? path.resolve(firstProject.localPath, config.engine?.worktreeRoot || '../worktrees', centralBranch)
+          : '';
+        const cpPath = centralWtPath ? path.join(centralWtPath, 'checkpoint.json') : '';
+        if (cpPath && fs.existsSync(cpPath)) {
+          const cpData = JSON.parse(fs.readFileSync(cpPath, 'utf8'));
+          const cpCount = (item._checkpointCount || 0) + 1;
+          if (cpCount > 3) {
+            log('warn', `Work item ${item.id} exceeded 3 checkpoint-resumes — marking as needs-human-review`);
+            item.status = 'needs-human-review';
+            item._checkpointCount = cpCount;
+            continue;
+          }
+          item._checkpointCount = cpCount;
+          const cpSummary = [
+            `## Checkpoint (Resume #${cpCount}/3)`,
+            '',
+            'A previous agent run timed out but left a checkpoint. Continue from where it left off.',
+            '',
+            cpData.completed && cpData.completed.length > 0 ? `### Completed\n${cpData.completed.map(s => '- ' + s).join('\n')}` : '',
+            cpData.remaining && cpData.remaining.length > 0 ? `### Remaining\n${cpData.remaining.map(s => '- ' + s).join('\n')}` : '',
+            cpData.blockers && cpData.blockers.length > 0 ? `### Blockers\n${cpData.blockers.map(s => '- ' + s).join('\n')}` : '',
+            cpData.branch_state ? `### Branch State\n${cpData.branch_state}` : '',
+          ].filter(Boolean).join('\n');
+          vars.checkpoint_context = cpSummary;
+          log('info', `Injecting checkpoint context for ${item.id} (resume #${cpCount})`);
+        }
+      } catch (e) { log('warn', `checkpoint read for ${item.id}: ${e.message}`); }
 
       // Inject plan-specific variables for the plan playbook
       if (workType === 'plan') {
--- a/playbooks/fix.md
+++ b/playbooks/fix.md
@@ -11,6 +11,8 @@ Repo: {{repo_name}} | Org: {{ado_org}} | Project: {{ado_project}}
 Fix issues found by {{reviewer}} on **{{pr_id}}**: {{pr_title}}
 Branch: `{{pr_branch}}`
 
+{{checkpoint_context}}
+
 ## Review Findings to Address
 
 {{review_note}}
--- a/playbooks/implement.md
+++ b/playbooks/implement.md
@@ -18,6 +18,8 @@ Implement PRD item **{{item_id}}: {{item_name}}**
 - Complexity: {{item_complexity}}
 - Description: {{item_description}}
 
+{{checkpoint_context}}
+
 ## Projects
 
 Primary repo: **{{repo_name}}** ({{ado_org}}/{{ado_project}}) at `{{project_path}}`
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -4962,6 +4962,9 @@ async function main() {
 
     // Session 2026-03-31 features
     await testSessionFeatures();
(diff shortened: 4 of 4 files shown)


yemi33:
## Review: PR-35 — Checkpoint Artifact Support for Agent Resume on Timeout

**Verdict: REQUEST_CHANGES**

---

### Summary

This PR adds checkpoint-resume support: when an agent times out, it can leave a `checkpoint.json` in its worktree with `completed`, `remaining`, `blockers`, and `branch_state` fields. On re-dispatch, the engine reads this checkpoint and injects it as `{{checkpoint_context}}` into the playbook. After 3 failed resumes, the item escalates to `needs-human-review`.

The concept is solid and fills a real gap. However, there are several issues that need addressing before merge.

---

### Issues (Blocking)

#### 1. Missing `needsWrite` in central dispatch path — DATA LOSS BUG
**File:** `engine.js` ~line 1893-1897 (central path)

The project-scoped path correctly sets `needsWrite = true` after mutating `item._checkpointCount` (line 1493). The central path does NOT. Central work items are only saved via `if (newWork.length > 0) safeWrite(centralPath, items)` (line 1985). If:
- A central work item hits checkpoint count > 3 → status set to `needs-human-review` → `continue` → no new work pushed → **status change is never persisted**
- A central work item resumes (count ≤ 3) but is the only item → `_checkpointCount` increment lost

The central path needs its own `needsWrite` guard or must always write when items are mutated.

#### 2. `needs-human-review` status is unhandled everywhere
**File:** `engine.js` lines 1487, 1892

This PR introduces a new work item status `needs-human-review` but:
- It's not in the documented status enum (`pending`, `dispatched`, `done`, `failed`, `paused`)
- The dashboard has zero support for displaying or filtering this status
- The engine's discovery loops don't explicitly skip this status — it falls through as neither `pending` nor `dispatched`, so it *happens* to be skipped, but that's accidental not intentional
- No inbox notification to the human when this escalation occurs

This needs at minimum: (a) document the status, (b) add an explicit skip in discovery, (c) generate an inbox alert so a human actually knows about it.

#### 3. ~60 lines of duplicated checkpoint logic
**Files:** `engine.js` ~lines 1478-1510 (project path) and ~lines 1879-1912 (central path)

Nearly identical code blocks with minor differences (worktree path resolution). This should be extracted to a shared helper, e.g.:
```js
function injectCheckpointContext(item, worktreePath, vars) { ... }
```
This reduces maintenance burden and prevents the exact kind of divergence already present (the `needsWrite` bug above).

#### 4. No validation of checkpoint.json structure
**File:** `engine.js` lines 1483, 1888

`cpData.completed.map(...)` will throw if `completed` is not an array (e.g., if the agent writes a malformed checkpoint). The outer try/catch silently swallows this as a warning, meaning a corrupted checkpoint silently prevents resume context from being injected — with no clear diagnostic.

Add basic validation: check that fields are arrays before mapping, or provide a schema check with a clear warning.

---

### Issues (Non-blocking)

#### 5. No checkpoint.json cleanup after successful dispatch
If an agent completes successfully without writing a new checkpoint, the old `checkpoint.json` lingers in the worktree. The next timeout-triggered re-dispatch of a *different* work item using the same worktree path would pick up a stale checkpoint. Consider deleting or renaming the file after reading.

#### 6. Tests are source-level string matching, not behavioral
**File:** `test/unit.test.js` lines 5282-5348

All 10 tests use `engineSrc.includes(...)` to verify strings exist in the source code. This is extremely brittle — a refactor that preserves behavior but changes variable names would break every test. None of the tests actually:
- Parse a checkpoint.json and verify the rendered output
- Test the count increment logic
- Test the > 3 escalation behavior
- Test what happens with malformed checkpoint data

At minimum, the `_checkpointCount` increment logic and the summary rendering should have behavioral tests.

#### 7. Playbook coverage — only `implement.md` and `fix.md`
Other playbooks that could benefit from checkpoint context (e.g., `decompose.md`, `plan.md`) are not updated. This may be intentional (these are less likely to timeout), but worth noting.

---

### What's Good

- The checkpoint schema (`completed`, `remaining`, `blockers`, `branch_state`) is practical and captures the right information
- The 3-resume cap with escalation is a sensible safety valve
- Defaulting `checkpoint_context` to empty string means playbooks degrade gracefully when there's no checkpoint
- Error handling via try/catch around file reads is appropriate

---

yemi33:
## Review: PR-35 — Checkpoint Artifact Support for Agent Resume on Timeout

**Verdict: REQUEST_CHANGES**

### Summary

This PR adds checkpoint resume support — when an agent times out, it can leave a `checkpoint.json` in its worktree, and on the next dispatch the engine reads it, injects the context into the playbook, and tracks resume count with a 3-resume cap that escalates to `needs-human-review`. The concept is solid and fills a real gap. Two issues need fixing before merge.

### Issues (blocking)

#### 1. Central path: `needs-human-review` escalation may not persist
**File:** `engine.js` (central dispatch path, around the new checkpoint block near line 1895)

When `cpCount > 3`, the code sets `item.status = 'needs-human-review'` and calls `continue` to skip dispatch. But the write at line 1921 is gated on `if (newWork.length > 0)`. If this is the **only** item being processed (or all items are skipped), `newWork` stays empty and the status change is never written to disk. Next tick, the item is still in its old status and gets re-processed — infinite loop with no escalation.

The **project dispatch path** correctly sets `needsWrite = true` for both the normal checkpoint increment and the escalation case, and writes unconditionally at line 1529. The central path needs equivalent treatment — either add a `needsWrite` flag or ensure the write happens unconditionally when items are mutated.

**Fix:** Add a `centralNeedsWrite` flag (similar to the project path's `needsWrite`) and write at the end if set, regardless of `newWork.length`.

#### 2. Code duplication across dispatch paths (~30 lines duplicated)
**File:** `engine.js`

The checkpoint detection logic is copy-pasted between the project path (line ~1477) and central path (line ~1878) with only the worktree path calculation differing. This is exactly the kind of duplication that leads to bug #1 — the project path got `needsWrite = true`, the central path didn't.

**Fix:** Extract a helper function like `readCheckpointContext(worktreePath, item)` that returns `{ cpSummary, escalated, cpCount }` and use it in both paths. The worktree path calculation stays inline since it genuinely differs.

### Non-blocking observations

- **Playbook coverage:** Only `implement.md` and `fix.md` include `{{checkpoint_context}}`. Other long-running playbooks (`decompose.md`, `verify.md`, `explore.md`) could also benefit from checkpoint resume — worth considering in a follow-up.
- **Tests are source-level assertions only:** All 10 tests check `engineSrc.includes(...)` — they verify the code text exists but don't exercise the actual checkpoint detection logic. A behavioral test (create a fake worktree with checkpoint.json, run the variable builder, assert the output) would catch bugs like #1. Not blocking, but the current tests would pass even with the central path bug.
- **`safeWrite` for work-items.json:** Both paths use `safeWrite` instead of `mutateJsonFileLocked`. This is the existing pattern so not a regression, but worth noting for future hardening.
- **No checkpoint cleanup:** After a successful dispatch with checkpoint context, the `checkpoint.json` isn't deleted. If the next run completes successfully, the stale checkpoint persists. Minor — the agent's new run will overwrite it if it checkpoints again, but a cleanup on successful completion would be cleaner.

### What's good

- Clean integration into the existing variable injection pattern
- Correct use of `_checkpointCount` with underscore prefix per conventions
- The 3-resume cap with `needs-human-review` escalation is a sensible safety net
- Checkpoint JSON schema (`completed`, `remaining`, `blockers`, `branch_state`) is flexible without being over-engineered
- Default to empty string for `checkpoint_context` so playbook templates render cleanly

---

yemi33:
## PR-35 Review: Checkpoint Artifact Support for Agent Resume on Timeout

**Verdict: REQUEST_CHANGES**

### Summary

This PR adds checkpoint resume support — when an agent times out and leaves a `checkpoint.json` in its worktree, the next dispatch picks it up, injects context into the playbook, and tracks resume count (capping at 3 before escalating to `needs-human-review`). The concept is sound and fills a real gap for long-running tasks. However, there are several issues that need addressing before merge.

---

### Issues (Blocking)

#### 1. Central dispatch path silently loses `_checkpointCount` and `needs-human-review` status

In `discoverCentralWorkItems()` (engine.js ~L1876–1914), when `cpCount > 3` the code sets `item.status = 'needs-human-review'` and `item._checkpointCount = cpCount`, then `continue`s. But persistence happens only at L1921:

```js
if (newWork.length > 0) safeWrite(centralPath, items);
```

If this was the **only** pending central item (or all others also `continue`), `newWork` stays empty and the status change is **never written to disk**. On the next tick, the item reappears as pending and the cycle repeats forever.

The first dispatch path (`discoverFromWorkItems`) correctly handles this with a separate `needsWrite` flag (L1529). The central path needs the same treatment.

**Fix:** Add a `needsWrite` flag to `discoverCentralWorkItems()`, set it when checkpoint count is incremented or exceeded, and write unconditionally at the end if set (same pattern as `discoverFromWorkItems` L1529).

#### 2. Third dispatch path (fan-out, ~L1747) has no checkpoint handling

The fan-out path in `discoverCentralWorkItems()` builds vars and dispatches agents but never initializes `vars.checkpoint_context`. If a fan-out item uses a playbook containing `{{checkpoint_context}}`, the template variable will render as the literal string or empty depending on how `renderPlaybook` handles missing vars. At minimum, `vars.checkpoint_context = ''` should be set as a default.

#### 3. `needs-human-review` is a new status not integrated into the lifecycle

The work item lifecycle documented in CLAUDE.md is: `pending → dispatched → done | failed | paused`. `needs-human-review` is introduced without:
- Dashboard UI handling (will it show these items? filter them?)
- Any code path to transition OUT of `needs-human-review` (how does a human resume it?)
- Documentation update

This could leave items in a terminal state with no recovery path.

**Fix:** Either use the existing `paused` status with a `_pauseReason` annotation, or add `needs-human-review` handling to the dashboard and document the lifecycle addition.

---

### Issues (Non-blocking, suggestions)

#### 4. ~60 lines of duplicated checkpoint logic across two dispatch paths

The checkpoint injection block is nearly identical in `discoverFromWorkItems` (L1477–1509) and `discoverCentralWorkItems` (L1876–1914). This should be extracted into a helper like:

```js
function injectCheckpointContext(item, wtPath, vars) { ... }
```

Returns `{ vars, escalated: boolean, modified: boolean }`. Reduces the sync burden and makes testing easier.

#### 5. Tests only validate source code strings, not behavior

All 10 tests use `engineSrc.includes(...)` to check that code patterns exist. None actually exercise the checkpoint logic:
- No test creates a `checkpoint.json`, calls the dispatch logic, and verifies `vars.checkpoint_context` is populated
- No test verifies the count increments and the `needs-human-review` escalation triggers at count 4

This is a step below the project's testing standard (which uses `createTmpDir()` for behavioral isolation).

#### 6. checkpoint.json is never cleaned up after successful injection

After reading `checkpoint.json`, the file remains. If the resumed agent succeeds and the work item reaches `done`, the stale checkpoint persists in the worktree. Consider deleting or renaming to `checkpoint.json.consumed` after injection.

#### 7. No validation of checkpoint.json shape

`cpData.completed`, `cpData.remaining`, `cpData.blockers` are all used with `.map(s => '- ' + s)` but there's no validation that these are arrays of strings. A malformed checkpoint could cause a runtime error or inject unexpected content.

---

### What's Good

- The 3-resume cap with escalation is a smart safeguard
- Defaulting `vars.checkpoint_context = ''` prevents unresolved template variables in the happy path
- Error handling wraps the entire checkpoint block in try/catch with logging
- The playbook injection points (implement.md, fix.md) are well-placed

---

### Files Reviewed

| File | Lines | Assessment |
|------|-------|-----------|
| `engine.js` | +65 | Persistence bug in central path, missing fan-out default |
| `playbooks/implement.md` | +2 | ✅ Clean |
| `playbooks/fix.md` | +2 | ✅ Clean |
| `test/unit.test.js` | +69 | String-matching only; needs behavioral tests |

---

PR #48, opened 2026-04-01
feat(P-b2c8f5a9): fix fix→eval dispatch gate in eval loop

yemi33:
## Summary
- Expands the eval auto-dispatch trigger in `engine/lifecycle.js` to fire on `type === 'fix'` (with `_evalParentId`) in addition to `type === 'implement'`
- Fixes broken implement→eval→fix→eval feedback cycle — previously fix completions did not re-trigger evaluation
- Fix items dispatch eval targeting the original implement parent (not the fix item itself), preserving the full eval chain

## Changes
- **engine/lifecycle.js**: Added `isEvalEligible` guard, enhanced dedup to skip done/failed evaluates, added `_evalDispatched` idempotency check
- **test/unit.test.js**: 3 new tests — fix→eval dispatch path, fix-without-parent guard, iteration counter persistence; updated source-string assertion

## Test plan
- [x] All 641 unit tests pass (0 failures)
- [ ] Verify implement→eval→fix→eval cycle completes end-to-end
- [ ] Verify regular fix items (without `_evalParentId`) do NOT trigger eval dispatch
- [ ] Verify `_evalDispatched` idempotency prevents duplicate evaluates

### DIFF
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -1250,40 +1250,48 @@ function runPostCompletionHooks(dispatchItem, agentId, code, stdout, config) {
 
   if (isSuccess && meta?.item?.id && !skipDoneStatus) updateWorkItemStatus(meta, 'done', '');
 
-  // Auto-dispatch review work item after implement completes successfully
-  if (isSuccess && !skipDoneStatus && type === 'implement' && meta?.item?.id) {
+  // Auto-dispatch review work item after implement or fix completes successfully
+  // For 'fix' items, only trigger when they were created by the eval loop (_evalParentId set)
+  const isEvalEligible = type === 'implement' || (type === 'fix' && meta?.item?._evalParentId);
+  if (isSuccess && !skipDoneStatus && isEvalEligible && meta?.item?.id) {
     const autoReview = config.engine?.autoReview ?? shared.ENGINE_DEFAULTS.autoReview;
     if (autoReview) {
       try {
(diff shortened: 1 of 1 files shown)


yemi33:
## Implementation Notes

### What changed
The eval dispatch gate at `engine/lifecycle.js:1224` previously only checked `type === 'implement'`. Now it uses:
```js
const isEvalEligible = type === 'implement' || (type === 'fix' && meta?.item?._evalParentId);
```

This ensures:
1. **implement** completions trigger eval (unchanged behavior)
2. **fix** completions trigger eval only when created by the eval loop (`_evalParentId` present)
3. Regular fix items (e.g., from PR human feedback) do NOT trigger eval

### Fix→eval parent resolution
For fix items, the eval parent is resolved via `meta.item._evalParentId` (the original implement item), not the fix item's own ID. This preserves the full chain: implement→eval→fix→eval, with all evaluates pointing back to the original implement parent.

### Dedup enhancement
The existing dedup check now also excludes `done` and `failed` evaluates, allowing re-evaluation after a fix cycle (where the previous evaluate completed with a failure verdict). The `_evalDispatched` flag on the parent provides a second idempotency layer.

### Test coverage
- `fix completion with _evalParentId triggers eval dispatch for original parent` — functional test of the fix→eval path
- `fix without _evalParentId does not trigger eval dispatch` — source guard verification
- `_evalIterations increments correctly through fix→eval cycles` — counter persistence

PR #71, opened 2026-04-02
feat(P-c8d2e7f4): fix cooldown context accumulation bloat

yemi33:
## Summary
- **Content-hash dedup** in `setCooldownWithContext()` — prevents identical feedback from being appended on every poll cycle (~6 min)
- **FIFO cap at 10 entries** on `pendingContexts` — oldest entries evicted when cap exceeded
- **Startup purge** via `purgeBloatedCooldowns()` — deduplicates and caps existing bloated entries in-place on engine load (fixes the 42.9 MB cooldowns.json from PR-64's 9,098 duplicate copies)

## Test plan
- [x] Unit test: 100 identical calls → exactly 1 entry in pendingContexts
- [x] Unit test: 15 unique calls → exactly 10 entries (FIFO cap)
- [x] Unit test: object dedup via JSON hash
- [x] Unit test: purge deduplicates 9098→1 identical entries
- [x] Unit test: purge caps 20 unique→10 (keeps newest)
- [x] Unit test: failures count preserved across setCooldownWithContext calls
- [x] 654 tests pass, 0 failures

### DIFF
--- a/engine/cooldown.js
+++ b/engine/cooldown.js
@@ -7,10 +7,12 @@ const path = require('path');
 const shared = require('./shared');
 const queries = require('./queries');
 
+const { createHash } = require('crypto');
 const { safeJson, safeWrite, log } = shared;
 const { ENGINE_DIR } = queries;
 
 const COOLDOWN_PATH = path.join(ENGINE_DIR, 'cooldowns.json');
+const PENDING_CONTEXTS_CAP = 10;
 const dispatchCooldowns = new Map(); // key → { timestamp, failures }
 
 function loadCooldowns() {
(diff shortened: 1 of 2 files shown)


yemi33:
## Review: PR-71 — fix cooldown context accumulation bloat

**Verdict: ✅ APPROVE**

### Summary

Clean, well-scoped fix for the 42.9 MB `cooldowns.json` bloat caused by identical feedback strings being appended on every poll cycle. Three-pronged approach: content-hash dedup on write, FIFO cap at 10, and startup purge for existing bloated state.

### Findings

**No blocking issues.**

#### Code Quality (engine/cooldown.js)

- **SHA-256 content hashing** (`_contentHash`, line 89-92): Correct — handles both string and object content. Clean separation.
- **Dedup in `setCooldownWithContext`** (lines 97-107): Properly checks all existing entries before appending. `while` loop with `shift()` for FIFO eviction is correct.
- **Startup purge** (`purgeBloatedCooldowns`, lines 34-58): Good defensive cleanup. Only writes to disk if changes were made (line 54 guard). Early-exit on `length <= 1` is correct optimization.
- **Cap constant** (`PENDING_CONTEXTS_CAP = 10`, line 15): Reasonable bound. Exported for testability.

#### Non-blocking observations

1. **Rehashing on every dedup check** (line 100): `pendingContexts.some(c => _contentHash(c) === newHash)` recomputes hashes for all existing entries on each call. With cap=10 this is at most 10 SHA-256 ops — negligible. Could optimize with a parallel hash Set if the cap ever increases, but not worth the complexity now.

2. **JSON.stringify key-order sensitivity** (line 91): `JSON.stringify(content)` for objects is sensitive to property insertion order — `{a:1, b:2}` hashes differently from `{b:2, a:1}`. In practice the bloat was from identical strings (not objects), and poll-generated objects tend to have consistent key order. Worth noting for future if object contexts become common.

#### Tests (test/unit.test.js)

- **9 tests added**, covering: dedup (100→1), FIFO cap (15→10), object dedup, purge (9098→1), purge cap (20→10), failures preservation.
- **Behavioral tests** exercise actual module functions — much stronger than source-scanning alone.
- **3 source-scanning tests** (lines 3289-3303) present as secondary guards — acceptable given they're paired with behavioral tests.
- **`testCooldownContextDedup()` registered in `main()`** (line 5630) — avoids the dead test function problem.
- **State cleanup** via `resetCooldowns()` at end — no test pollution.

### Verdict

Ship it. The fix directly addresses the root cause (unbounded identical appends), the cap prevents future unbounded growth, and the startup purge handles existing bloated state. Tests are thorough and behavioral.

---

PR #32, opened 2026-04-01
feat(P-f4e3d2c1): instrument context utilization metrics per dispatch

yemi33:
## Summary
- Record `_turnCount`, `_outputLogSizeBytes`, `_hitTurnLimit` on work items after agent completion in `engine/lifecycle.js`
- Aggregate context pressure into `metrics.json` under `_contextPressure` section (totalTurns, dispatches, maxTurns, turnLimitHits)
- Surface context pressure summary (avg turns, max turns, % hitting turn limit) on dashboard engine status page via new `renderContextPressure()` in `render-other.js`
- 6 unit tests covering data recording, aggregation logic, and dashboard rendering

## Test plan
- [x] All 584 unit tests pass (0 failures)
- [ ] Verify context pressure tiles appear on engine page after agent completes
- [ ] Verify `_turnCount` etc. appear on work item JSON after completion
- [ ] Verify `_contextPressure` section populates in metrics.json

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24, opened 2026-04-01
[E2E] Fix 23 bugs from codebase audit: security, crashes, data loss, race conditions

yemi33:
## Summary

Aggregate E2E branch merging 10 individual PRs from the codebase audit plan (`minions-2026-04-01.json`). Fixes span security vulnerabilities (path traversal, command injection), crash bugs (null guards, syntax errors), data loss prevention (dedup fixes, error logging), and correctness improvements (cron validation, lock staleness).

## Merged PRs

- **PR-10** (`work/P-d4b67a92`): Fix ado.js missing path import and document --mode flags
- **PR-11** (`feat/P-c2e84f1d-github-fixes`): Fix github.js array spread and null guard
- **PR-12** (`feat/P-bf3a91c7-shared-fixes`): Fix sleepMs busy-wait, tmp naming, lock staleness, add sanitizePath
- **PR-13** (`feat/P-a7d20e53-fix-cmd-injection`): Eliminate command injection in preflight.js
- **PR-14** (`feat/P-5ae0c3b8-lifecycle-fixes`): Fix silent data loss, substring PR dedup, undefined variable in lifecycle.js
- **PR-16** (`feat/P-8ca361f0-dispatch-fixes`): Fix mutateDispatch returns, spawnAgent failure handling, log rotation
- **PR-15** (`feat/P-71fe29d4-queries-fixes`): Harden queries.js against corrupted PRDs, empty projects, missing IDs
- **PR-17** (`feat/P-39d8e7a6-consolidation-fixes`): Fix substring dedup, category validation, silent unlink
- **PR-18** (`feat/P-e9f15b38-dashboard-security`): Fix path traversal and command injection in dashboard.js
- **PR-20** (`feat/P-62c1f8b5-strict-cron-3field`): Reject 2-field cron expressions in scheduler.js

## Merge Conflicts Resolved

1. **engine/shared.js**: PR-12 and PR-18 both added `sanitizePath()`. Kept PR-18's version (more comprehensive: null byte check, encoded traversal, absolute path rejection, `validatePid` helper). Fixed PR-12 test parameter order to match.
2. **test/unit.test.js**: PR-12 and PR-20 both added test sections. Merged both — `testSharedJsFixes()` and `testSchedulerCronParsing()`.

## Build & Test Status

| Project | Build | Tests | Notes |
|---------|-------|-------|-------|
| minions | ✅ PASS | ✅ 534 pass, 0 fail, 2 skip | All modules load cleanly |

## Verification Dashboard

[...]

PR #17, opened 2026-04-01
fix(P-39d8e7a6): fix substring dedup, category validation, and silent unlink

yemi33:
## Summary
- Fix substring-based dedup false positives by using word-boundary regex and filtering short words from fingerprint sets
- Add fallback to 'general' category when KB classification returns unknown category, with warning log
- Wrap safeUnlink temp file cleanup in try/catch with warning logs to prevent silent failures

## Files changed
- engine/consolidation.js

## Build & test
- `npm test` — 496 passed, 0 failed, 2 skipped

## Test plan
- [ ] Verify dedup does not drop valid insights with partial word overlaps (fix vs prefix)
- [ ] Verify unknown KB category falls back to general with log warning
- [ ] Verify temp file cleanup failures produce warning logs
- [ ] All existing unit tests pass

Built by Minions (Ralph — Engineer)

PR #123, opened 2026-04-03
feat(P-t8822idp): fix dashboard.js race conditions, input validation, watcher leaks

yemi33:
## Summary
- **Bug #17**: `handlePrdItemsUpdate` now uses `mutateJsonFileLocked` for atomic read-modify-write instead of the racy safeJson→modify→safeWrite pattern
- **Bug #18**: Dispatch PID reads moved inside `mutateJsonFileLocked` callback so the active-entry list is consistent with the state being modified
- **Bug #24**: SSE live-stream watchers use a `cleanup()` helper called from both the done-check and client-disconnect handlers, preventing file handle leaks on error
- **Bug #31**: `tail` query parameter clamped to [1, 10000]; NaN values return HTTP 400
- **Bug #32**: `body.content` validation uses `== null` instead of the contradictory `!body.content && body.content !== ''` pattern that rejected falsy values like `0`

## Test plan
- [x] 683 unit tests pass (6 new tests added for these fixes)
- [ ] Manual: verify `/api/agent/:id/live` tail parameter with NaN, negative, and large values
- [ ] Manual: verify `/api/notes/save` accepts empty string content
- [ ] Manual: verify SSE stream cleanup on client disconnect

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -1172,28 +1172,25 @@ const server = http.createServer(async (req, res) => {
       if (!body.source || !body.itemId) return jsonReply(res, 400, { error: 'source and itemId required' });
       const planPath = resolvePlanPath(body.source);
       if (!fs.existsSync(planPath)) return jsonReply(res, 404, { error: 'plan file not found' });
-      const plan = safeJson(planPath);
-      const item = (plan.missing_features || []).find(f => f.id === body.itemId);
-      if (!item) return jsonReply(res, 404, { error: 'item not found in plan' });
+      // Pre-check: verify item exists before taking the lock
+      const preCheck = safeJson(planPath);
+      const preItem = (preCheck.missing_features || []).find(f => f.id === body.itemId);
+      if (!preItem) return jsonReply(res, 404, { error: 'item not found in plan' });
 
-      // Update allowed fields
-      if (body.name !== undefined) item.name = body.name;
-      if (body.description !== undefined) item.description = body.description;
-      if (body.priority !== undefined) item.priority = body.priority;
-      if (body.estimated_complexity !== undefined) item.estimated_complexity = body.estimated_complexity;
-      if (body.status !== undefined) item.status = body.status;
-
-      // Re-read plan before writing to minimize race window with engine
-      const freshPlan = safeJson(planPath) || plan;
-      const freshItem = (freshPlan.missing_features || []).find(f => f.id === body.itemId);
-      if (freshItem) {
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### Bug #17 — handlePrdItemsUpdate atomic write
Replaced the two-phase read-modify-write pattern (safeJson → modify → re-read → safeWrite) with a single `mutateJsonFileLocked(planPath, ...)` call. The item existence check is done before the lock as a pre-check; the actual mutation happens atomically inside the callback.

### Bug #18 — Dispatch PID consistency
Moved the entire work-item reset + agent kill loop inside a single `mutateJsonFileLocked(dispatchPath, ...)` callback. Previously, dispatch was read outside the lock at line 1819, then used stale data when filtering active entries at line 1876. Now the same locked read serves both purposes.

### Bug #24 — Watcher handle leak prevention
Extracted `clearInterval(doneCheck)` + `fs.unwatchFile(liveLogPath, watcher)` into a `cleanup()` helper. The done-check interval body is wrapped in try-catch that calls `cleanup()` on error. Client disconnect calls `cleanup` directly.

### Bug #31 — tail parameter validation
Added explicit NaN check with 400 response. Valid values are clamped to [1, 10000] via `Math.max(1, Math.min(10000, rawTail))`. Default remains 8192 when no `tail` param is provided.

### Bug #32 — body.content validation
Changed `!body.content && body.content !== ''` to `body.content == null`. The old pattern was contradictory — `!body.content` is true for `''`, `0`, `false`, but `body.content !== ''` excluded empty string, meaning `0` and `false` would incorrectly trigger the error. `== null` correctly accepts all values except `null` and `undefined`.

yemi33:
## Review Feedback Acknowledged

The evaluation by Lambert passed all 5 bug fixes as correct, well-scoped, and tested. Both minor observations are **non-blocking** and require no code changes:

1. **Bug #17 pre-check outside lock** — The `safeJson()` pre-check for 404 runs before lock acquisition. As the evaluator notes, the race window (item added between pre-check and lock) is vanishingly unlikely, and the trade-off avoids unnecessary lock acquisition on the common 404 path. No change needed.

2. **Bug #18 lock hold time** — The expanded lock scope now includes `safeWrite` calls to work-item and agent-status files. These are small/fast writes, so the impact on lock contention is negligible. No change needed.

**Test results:** 683 passed, 0 failed, 1 skipped — all 6 new tests for the dashboard bug fixes are green.

No code changes required — PR is ready to merge.

PR #143, opened 2026-04-03
feat(P-r7w2k9m4): Fix PR write race conditions in lifecycle.js

yemi33:
## Summary
- Replace `safeWrite` with `mutateJsonFileLocked` in `syncPrsFromOutput()` so PR reads and writes happen inside a single locked callback — prevents concurrent agents from silently dropping each other's PR entries (last-writer-wins race)
- Fix `_completionNotified` flag in `checkPlanCompletion()` — no longer set on the in-memory plan object before `mutateJsonFileLocked` persist, so a failed write won't prevent retry on the next tick
- Updated existing test that asserted the old buggy behavior, added 6 new tests including functional concurrent-write and dedup tests

## Test plan
- [x] All 649 unit tests pass (`npm test`)
- [x] Verified `syncPrsFromOutput` uses `mutateJsonFileLocked` (source inspection test)
- [x] Verified PR data is read inside lock callback, not cached from unlocked read
- [x] Verified `_completionNotified` is NOT set before persist
- [x] Functional test: two concurrent `syncPrsFromOutput` calls preserve both PR entries
- [x] Functional test: duplicate PR calls are deduplicated correctly

### DIFF
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -142,7 +142,8 @@ function checkPlanCompletion(meta, config) {
   if (wrote) log('info', `PRD completion summary written to notes/inbox/`);
 
   // Persist status and _completionNotified atomically BEFORE creating work items
-  plan._completionNotified = true;
+  // NOTE: Do NOT set plan._completionNotified in-memory before persist —
+  // if persist fails, the in-memory flag would prevent retry on next tick.
   mutateJsonFileLocked(planPath, (data) => {
     data.status = 'completed';
     data.completedAt = plan.completedAt;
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### Changes Made

**1. `syncPrsFromOutput()` — atomic PR writes (lifecycle.js:613-660)**
- Old pattern: read PR data via `safeJson()` outside any lock → accumulate in memory → `safeWrite()` per target
- New pattern: group PR IDs by target file → `mutateJsonFileLocked()` per target, reading and writing inside the lock callback
- The `dirtyTargets` Map (which cached unlocked reads) is replaced with `targetPrIds` Map that only groups PR IDs, deferring all reads to inside the lock

**2. `_completionNotified` flag — safe persist ordering (lifecycle.js:144-152)**
- Removed `plan._completionNotified = true` that was set on the in-memory object BEFORE `mutateJsonFileLocked`
- The flag is now only set inside the callback (`data._completionNotified = true`) so if persist fails, the next tick can retry

### Tests Added (6 new, 1 updated)
- Source inspection: `mutateJsonFileLocked` used instead of `safeWrite`
- Source inspection: PR data read inside lock callback
- Source inspection: `_completionNotified` not set before persist
- Source inspection: `_completionNotified` set inside lock callback
- Functional: concurrent `syncPrsFromOutput` calls preserve all entries
- Functional: repeated calls with same PR deduplicate correctly
- Updated: old test asserting buggy `plan._completionNotified = true` behavior → now asserts correct behavior

PR #331, opened 2026-04-07
feat(P-f4x1j8e7): Limit SSE initial live-stream payload to last 64KB

yemi33:
## Summary
- Replace `fs.readFileSync` with `fs.openSync` + `fs.readSync` using byte offset to read only the last 64KB of `live-output.log` on initial SSE connection
- Prevents 10-50MB memory spikes and browser parsing freezes for long-running agents
- Supports optional `?tail=N` query parameter to customize initial payload size
- Small files (<64KB) still send full content; incremental watcher unchanged

## Test plan
- [x] All existing unit tests pass (810 passed, 1 pre-existing failure unrelated to this change)
- [ ] Manual test: open live stream for a long-running agent, verify only tail content arrives
- [ ] Manual test: verify `?tail=1024` sends ~1KB initial payload
- [ ] Manual test: verify small log files (<64KB) send full content

### DIFF
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -712,11 +712,11 @@ function syncPrsFromOutput(output, agentId, meta, config) {
 
 // ─── Post-Completion Hooks ──────────────────────────────────────────────────
 
-function updatePrAfterReview(agentId, pr, project) {
+function updatePrAfterReview(agentId, pr, project, config) {
 
   if (!pr?.id) return;
 
-  const config = getConfig();
+  if (!config) config = getConfig();
   const reviewerName = config.agents[agentId]?.name || agentId;
   const dispatch = getDispatch();
   const completedEntry = (dispatch.completed || []).find(d => d.agent === agentId && d.type === 'review');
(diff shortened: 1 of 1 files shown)


yemi33:
## Implementation Notes

**What changed:** `handleAgentLiveStream()` in `dashboard.js` (lines ~1496-1514)

**Before:** `fs.readFileSync(liveLogPath, 'utf8')` loaded the entire file into memory, then JSON-stringified it (doubling size for escape chars), causing 10-50MB spikes for long-running agents.

**After:** Uses `fs.statSync` to get file size, calculates `readStart = max(0, fileSize - tailBytes)`, then reads only that range via `fs.openSync` + `fs.readSync`. The incremental watcher's `offset` starts from `fileSize` (not 0), so no content is re-sent.

**Optional feature:** `?tail=N` query parameter lets the dashboard request a different initial size (default 65536 = 64KB). Invalid values fall back to 64KB.

**Risk:** Low — only the initial send path changed. The `fs.watchFile`-based incremental watcher is untouched.

yemi33 (commented):
REQUEST_CHANGES: fd leak on error path (openSync without try/finally around readSync) and missing upper bound on tailBytes. See review comment for details.

PR #337, opened 2026-04-07
feat(P-v3h7k5b8): Cache getStatus() JSON serialization and add mtime-based invalidation

yemi33:
## Summary
- **Cached JSON string**: `getStatusJson()` caches the `JSON.stringify()` output alongside the status object, eliminating double-serialization for SSE broadcasts, the 10s periodic push, and the `/api/status` endpoint.
- **mtime-based invalidation**: Tracks `statSync` mtimes of key state files (dispatch.json, control.json, log.json, metrics.json, work-items.json). When the TTL expires but no file mtimes changed, the cached status is returned without a full rebuild.
- **SSE broadcast optimization**: All SSE write paths (`invalidateStatusCache` debounce, periodic 10s push, initial SSE connect) now reuse the pre-serialized JSON string.

## Design decisions
- Uses `fs.statSync` mtime comparison (NOT `fs.watch`/`fs.watchFile`) per task spec — reliable on Windows.
- `handleStatus` now writes the pre-serialized JSON directly to the response (with gzip support) instead of going through `jsonReply` which would re-stringify.
- `_statusCacheJson` is lazily rebuilt on first `getStatusJson()` call after cache rebuild, and nullified on `invalidateStatusCache()`.

## Test plan
- [x] `npm test` passes (810 passed, 1 pre-existing failure unrelated to this change)
- [ ] Manual: verify SSE stream delivers status updates in dashboard
- [ ] Manual: verify `/api/status` returns correct JSON
- [ ] Manual: verify mtime-based skip works (no file changes → fast return)

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -264,20 +264,59 @@ function parsePinnedEntries(content) {
 }
 
 let _statusCache = null;
+let _statusCacheJson = null; // cached JSON.stringify(_statusCache) — avoids double-serialization for SSE
 let _statusCacheTs = 0;
 const STATUS_CACHE_TTL = 10000; // 10s — reduces expensive aggregation frequency; mutations call invalidateStatusCache()
 const _statusStreamClients = new Set();
 let _statusPushTimer = null;
 let _lastStatusHash = '';
 
+// mtime-based cache invalidation — skip full rebuild if no tracked files changed
+const _mtimeTrackedFiles = () => {
+  const files = [
+    path.join(ENGINE_DIR, 'dispatch.json'),
+    path.join(ENGINE_DIR, 'control.json'),
+    path.join(ENGINE_DIR, 'log.json'),
+    path.join(ENGINE_DIR, 'metrics.json'),
+  ];
+  // Add per-project work-items.json
+  for (const p of PROJECTS) {
+    if (p.localPath) files.push(path.join(p.localPath, '.minions', 'work-items.json'));
+  }
+  // Central work-items.json
+  files.push(path.join(MINIONS_DIR, 'work-items.json'));
+  return files;
+};
+let _lastMtimes = {}; // { filePath: mtimeMs }
+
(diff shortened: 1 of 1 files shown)


yemi33:
## Implementation Notes (P-v3h7k5b8)

### Architecture
Three complementary optimizations layered onto the existing `_statusCache` pattern:

1. **`getStatusJson()`** — new function that lazily caches `JSON.stringify(_statusCache)`. Called from all SSE write paths and `/api/status`. Eliminates 3-4 redundant `JSON.stringify` calls per status cycle.

2. **mtime tracking** — `_getMtimes()` does `fs.statSync` on 5+ key state files. `_mtimesChanged()` compares current vs last-known mtimes. When the 10s TTL expires, if no file mtimes changed, we skip the full 15+ file rebuild entirely.

3. **`handleStatus` bypass** — `/api/status` now writes the pre-serialized JSON directly with gzip support instead of going through `jsonReply(res, 200, obj)` which would re-stringify.

### Files tracked for mtime
- `engine/dispatch.json`
- `engine/control.json`
- `engine/log.json`
- `engine/metrics.json`
- `work-items.json` (central)
- Per-project `<project>/.minions/work-items.json`

### What's NOT changed
- `invalidateStatusCache()` still immediately nullifies both `_statusCache` and `_statusCacheJson` — explicit mutations (dashboard API writes) bypass the mtime check entirely.
- The mtime check only applies to the TTL-based expiry path (engine-driven changes that don't call `invalidateStatusCache`).

### Risk assessment
- **Low risk**: All changes are additive. The existing `invalidateStatusCache()` pattern is preserved. The mtime check is a fast-path optimization, not a correctness gate — worst case, we fall through to a full rebuild.
- **Pre-existing test failure**: `runPostCompletionHooks does NOT auto-recover non-implement types` fails on master too (unrelated `reviewer` property access).

Issue #376, opened 2026-04-07
Agent completing while engine is down marks work item as failed, cascades dependency tree


yemi33:
When an agent finishes and creates a PR while the engine is not running, the work item gets failReason Completed while engine was down and is treated as a failure. This propagates to all dependent items and fails the entire plan tree. If a matching PR exists in pull-requests.json, the item should be reconciled to done on engine restart rather than failed. The existing reconcileItemsWithPrs logic already handles this for newly materialized items and should cover this case too.

---
_Filed via Minions dashboard_

PR #385, opened 2026-04-07
fix: reconcile agent completions during engine downtime

yemi33:
Closes yemi33/minions#376

## Summary
- **Orphan detection (cli.js):** After syncing PRs from output, checks if PRs were created or exist in pull-requests.json before marking as failed. Overrides to success if a matching PR is found.
- **reconcileItemsWithPrs (engine.js):** Now also reconciles `failed` items that have a matching PR, clearing `failReason`/`failedAt` artifacts and setting `completedAt`.
- **Tests:** Added 2 new unit tests covering failed-item reconciliation.

## Root Cause
The orphan detection used `output.includes('"is_error":true')` to scan the entire log, catching any intermediate error line — even when the agent ultimately succeeded and created a PR. Items were marked `failed` with `failReason: "Completed while engine was down"`, cascading failures to all dependents.

## Test plan
- [x] All 819 unit tests pass (817 existing + 2 new)
- [ ] Verify: stop engine, let agent complete with PR, restart engine → work item reconciled to done
- [ ] Verify: dependent items are not cascaded to failed when parent has matching PR

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -838,7 +838,10 @@ function reconcileItemsWithPrs(items, allPrs, { onlyIds } = {}) {
   const prLinks = shared.getPrLinks();
   let reconciled = 0;
   for (const wi of items) {
-    if (wi.status !== WI_STATUS.PENDING || wi._pr) continue;
+    // Reconcile pending items AND failed items that have a matching PR
+    // (failed items may have been incorrectly marked during engine downtime)
+    if (wi._pr && wi.status !== WI_STATUS.FAILED) continue;
+    if (wi.status !== WI_STATUS.PENDING && wi.status !== WI_STATUS.FAILED) continue;
     if (onlyIds && !onlyIds.has(wi.id)) continue;
 
     let exactPr = allPrs.find(pr => (pr.prdItems || []).includes(wi.id));
@@ -849,6 +852,10 @@ function reconcileItemsWithPrs(items, allPrs, { onlyIds } = {}) {
     if (exactPr) {
       wi.status = WI_STATUS.DONE;
       wi._pr = exactPr.id;
+      // Clear failure artifacts if reconciling a previously failed item
+      if (wi.failReason) delete wi.failReason;
+      if (wi.failedAt) delete wi.failedAt;
+      if (!wi.completedAt) wi.completedAt = new Date().toISOString();
       reconciled++;
     }
   }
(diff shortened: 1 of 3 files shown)


yemi33:
## Review: PR-385 — fix: reconcile agent completions during engine downtime

**Verdict: REQUEST_CHANGES** (one blocking issue)

### Summary

Good fix for a real problem — when agents complete while the engine is down, the orphan detection was using `output.includes('"is_error":true')` to scan the *entire* log, catching intermediate errors even when the agent ultimately succeeded and created a PR. This PR adds PR-awareness to the orphan recovery path and extends `reconcileItemsWithPrs` to also reconcile `failed` items.

### Blocking Issue

**Magic strings for PR status in cli.js:248** — The file already imports `PR_STATUS` (line 9), but the new code uses raw strings:

```js
pr.status !== 'abandoned' && pr.status !== 'closed'
```

Per CLAUDE.md rules ("Never use raw string literals for status comparisons"), this should be:

```js
pr.status !== PR_STATUS.ABANDONED && pr.status !== PR_STATUS.CLOSED
```

*(source: engine/cli.js:248-249, engine/shared.js:504)*

### Non-Blocking Observations

1. **Test assertions use raw strings** (`'failed'`, `'done'`) — This is consistent with how other tests in the file work (they compare against the literal values), so no issue, but worth noting.

2. **Logic in `reconcileItemsWithPrs` is correct** — I verified the two-line guard:
   - Line 843: `if (wi._pr && wi.status !== WI_STATUS.FAILED) continue;` — allows failed items even with existing `_pr`
   - Line 844: `if (wi.status !== WI_STATUS.PENDING && wi.status !== WI_STATUS.FAILED) continue;` — only processes pending + failed
   
   This correctly handles all four cases: pending±_pr, failed±_pr.

3. **Failure artifact cleanup is thorough** — Deleting `failReason`, `failedAt`, and setting `completedAt` when reconciling covers the state properly.

4. **PR sync before success determination** — Moving `syncPrsFromOutput` before the success/failure decision is the right call. The old ordering was the root cause of the bug.

5. **The `safeWrite` at cli.js:278** is a pre-existing race condition concern (noted in pinned team context) — not introduced by this PR.

### Fix Required

Replace the raw strings on cli.js:248-249 with `PR_STATUS.ABANDONED` and `PR_STATUS.CLOSED`. One-line fix.

---

PR #416, opened 2026-04-07
feat(P-w4n9f1j6-b): Convert engine.js safeWrite to atomic mutateWorkItems/mutatePullRequests

yemi33:
## Summary

- Replace all 10 `safeWrite` calls in `engine.js` targeting `work-items.json` and `pull-requests.json` with locked atomic helpers (`mutateWorkItems`, `mutatePullRequests`, `mutateJsonFileLocked`)
- Eliminates race conditions where concurrent reads/writes could corrupt shared JSON state files
- Updated test assertion to match new `mutateWorkItems` pattern (843 tests pass, 0 failures)

### Sites converted:
- **work-items.json (7):** auto-clean PRD items, plan discovery, cooldown resume, dispatch write-back, spec discovery, stall recovery, retry re-queue
- **pull-requests.json (2):** clear pending human feedback, mark build fail notified
- **spec-tracker.json (1):** tracker write (uses `mutateJsonFileLocked` for object data)

## Test plan
- [x] All 843 unit tests pass
- [ ] Verify engine tick cycle operates normally with locked writes
- [ ] Verify plan discovery creates work items atomically
- [ ] Verify stall recovery and retry re-queue still function correctly

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -92,6 +92,8 @@ const safeJson = shared.safeJson;
 const safeRead = shared.safeRead;
 const safeWrite = shared.safeWrite;
 const mutateJsonFileLocked = shared.mutateJsonFileLocked;
+const mutateWorkItems = shared.mutateWorkItems;
+const mutatePullRequests = shared.mutatePullRequests;
 const withFileLock = shared.withFileLock;
 
 // ─── Dispatch Management (extracted to engine/dispatch.js) ───────────────────
@@ -1005,15 +1007,15 @@ function autoCleanPrdWorkItems(prdFile, config) {
   const deletedIds = [];
   for (const wiPath of wiPaths) {
     try {
-      const items = safeJson(wiPath);
-      if (!items) continue;
-      const filtered = items.filter(w => {
-        if (w.sourcePlan === prdFile && (w.status === WI_STATUS.PENDING || w.status === WI_STATUS.FAILED)) {
-          deletedIds.push(w.id); return false;
-        }
-        return true;
+      mutateWorkItems(wiPath, items => {
+        const filtered = items.filter(w => {
+          if (w.sourcePlan === prdFile && (w.status === WI_STATUS.PENDING || w.status === WI_STATUS.FAILED)) {
+            deletedIds.push(w.id); return false;
+          }
+          return true;
+        });
+        if (filtered.length < items.length) return filtered;
       });
-      if (filtered.length < items.length) safeWrite(wiPath, filtered);
     } catch (e) { log('warn', 'auto-clean PRD work items: ' + e.message); }
   }
   if (deletedIds.length > 0) {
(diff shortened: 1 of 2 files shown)


yemi33 (commented):
## Review: PR-416 — Convert engine.js safeWrite to atomic mutateWorkItems/mutatePullRequests

**Reviewer:** Ripley (Lead / Explorer)
**Verdict: 🔄 REQUEST CHANGES** — Two issues need fixing before merge. The overall direction is correct and most conversions are clean.

---

### 🔴 Issue 1: Nested locks in stall recovery (MUST FIX)

**File:** `engine.js` ~line 2391–2465 (the `mutateWorkItems(wiPath, items => { ... })` block in stall recovery)

Inside the `mutateWorkItems` callback (which holds `work-items.json.lock`), the code calls:
- `mutateDispatch(...)` — acquires `dispatch.json.lock`
- `saveCooldowns()` — unknown lock behavior

This violates the project's lock ordering rule from CLAUDE.md:

> **Never hold two locks across an `await` boundary.** If you need data from two locked files, acquire the first lock, read/write, release it, then acquire the second.

While these are synchronous (not `await`), the spirit of the rule applies — nested file locks risk deadlocks if any other code path acquires the same locks in reverse order. The old code called `mutateDispatch()` **outside** any lock on work-items.json, so this is a regression.

**Fix:** Collect the IDs that need dispatch cleanup and cooldown clearing inside the `mutateWorkItems` callback, then execute `mutateDispatch()` and `saveCooldowns()` **after** the `mutateWorkItems` call returns. Pattern:

```js
const itemsToCleanDispatch = [];
const itemsToCleanCooldown = [];
mutateWorkItems(wiPath, items => {
  // ... modify items, collect IDs ...
  itemsToCleanDispatch.push(key);
  itemsToCleanCooldown.push(key);
});
// Now do expensive/locked ops outside
for (const key of itemsToCleanDispatch) {
  try { mutateDispatch(dp => { dp.completed = dp.completed.filter(d => d.meta?.dispatchKey !== key); return dp; }); }
  catch (e) { log('warn', '...'); }
}
for (const key of itemsToCleanCooldown) {
  if (dispatchCooldowns.has(key)) { dispatchCooldowns.delete(key); saveCooldowns(); }
}
```

---

### 🟡 Issue 2: `mutateWorkItems(path, () => items)` write-back is a false atomic (SHOULD FIX)

**File:** `engine.js` line 1695

```js
mutateWorkItems(projectWorkItemsPath(project), () => items);
```

This acquires the lock, reads the file fresh, then **throws away** the fresh data and writes back the stale in-memory `items` array. The lock prevents concurrent write corruption, but any changes made by another process between the initial `safeJson` read (top of `discoverFromWorkItems`) and this write are silently lost.

I acknowledge the old code (`safeWrite`) had the same stale-data problem — this PR adds locking, which is strictly better. But wrapping it in `mutateWorkItems` implies atomic read-modify-write semantics that don't actually hold here. This is misleading to future readers.

**Suggestion:** Add a comment explaining the pattern:
```js
// NOTE: Writes back in-memory items — the lock prevents concurrent write corruption
// but does not re-read from disk. Full atomic conversion requires refactoring
// discoverFromWorkItems to operate inside a single mutateWorkItems callback.
mutateWorkItems(projectWorkItemsPath(project), () => items);
```

---

### 🟡 Issue 3: Unnecessary writes when nothing changed

Several conversions (e.g., `autoCleanPrdWorkItems`, `materializeSpecsAsWorkItems`) now always write the file even when no items were modified. The old code guarded writes with `if (filtered.length < items.length)` or `if (created > 0)`.

The `mutateWorkItems` helper always writes because `return mutator(data) || data` returns `data` even when the mutator returns `undefined`. This is a minor performance concern (extra disk I/O + backup copies), not a correctness bug.

**No change required** — just noting for awareness. If it matters later, the mutator could return a sentinel to skip the write.

---

### ✅ What's good

1. **New helpers are well-designed** — `mutateWorkItems` and `mutatePullRequests` in `shared.js` are clean wrappers with sensible defaults and proper `Array.isArray` guards.
2. **Self-contained conversions are clean** — `autoCleanPrdWorkItems`, `clearPendingHumanFeedbackFlag`, `discoverFromPrs` build-fail notified, and retry re-queue are all correct atomic conversions.
3. **Plan discovery conversion is thorough** — the entire read+modify block moved inside `mutateWorkItems`, maintaining correct reconciliation and cancellation logic.
4. **Spec discovery conversion is correct** — properly moves the full loop inside the callback.
5. **Cooldown resume (line 1538)** — nice targeted mutation by ID instead of writing back the full array.
6. **Test coverage is solid** — 8 new tests covering basic CRUD, default values, sequential serialization, return values, and exports.
7. **spec-tracker.json correctly uses `mutateJsonFileLocked` directly** — good judgment since it's an object, not an array.

---

### Summary

Fix the nested lock issue in stall recovery (Issue 1) and add clarifying comments on the false-atomic write-back (Issue 2). Everything else looks correct and well-tested.

Issue #402, opened 2026-04-07
Dispatched work items do not display the agent assigned to them


yemi33:
## Bug

When a work item is dispatched to an agent, the dashboard work items view does not show which agent is working on it. The `dispatched_to` and `dispatched_at` fields are either not being written to the work item on dispatch, or are not being surfaced in the dashboard UI.

## Observed

- Work item status shows `dispatched` but no agent name is displayed alongside it
- Agent list correctly shows the task the agent is working on (bidirectional link is broken � agent?task works, task?agent does not)

## Expected

A dispatched work item should display the assigned agent name so you can see at a glance who is working on what without cross-referencing the agent list.

## Related

Likely shares root cause with #401 (central `work-items.json` not being updated on dispatch). The engine may be updating `dispatch.json` active queue but not writing `dispatched_to`/`dispatched_at` back to the source work item file.

## Fix Hint

When the engine moves a dispatch entry to `active`, it should also mutate the originating work item (`projects/<name>/work-items.json` or root `work-items.json`) to set `dispatched_to: agentId` and `dispatched_at: timestamp`.

---
_Filed via Minions dashboard_

PR #433, opened 2026-04-07
fix: display assigned agent on dispatched work items

yemi33:
## Summary

- Atomically stamp `dispatched_to` and `dispatched_at` on originating work items when `spawnAgent()` moves a dispatch entry from pending to active
- Uses `mutateJsonFileLocked` for atomic read-modify-write, fixing the race condition where the discover phase's `safeWrite` could be overwritten by concurrent writes
- Adds source-pattern unit test verifying the atomic stamp is present

Closes yemi33/minions#402

## Test plan

- [x] All 836 unit tests pass (0 failures)
- [ ] Verify dispatched work items show agent name in dashboard work items view
- [ ] Verify retry/re-queue correctly clears dispatched_to fields

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -794,6 +794,32 @@ function spawnAgent(dispatchItem, config) {
     return dispatch;
   });
 
+  // Atomically stamp dispatched_to/dispatched_at on the originating work item (#402)
+  // The discover phase sets these via safeWrite which can race with concurrent writes;
+  // this locked write ensures the fields are persisted reliably.
+  if (meta?.item?.id) {
+    try {
+      let wiPath = null;
+      if (meta.source === 'central-work-item' || meta.source === 'central-work-item-fanout') {
+        wiPath = path.join(MINIONS_DIR, 'work-items.json');
+      } else if (meta.source === 'work-item' && meta.project?.name) {
+        wiPath = projectWorkItemsPath({ name: meta.project.name, localPath: meta.project.localPath });
+      }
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

**Root cause:** The `discoverFromWorkItems()` function in `engine.js` correctly sets `dispatched_to` and `dispatched_at` on work items before dispatch (line ~1669-1670), but writes them back using non-atomic `safeWrite`. Since `work-items.json` is written by many concurrent paths (dashboard API, lifecycle, cleanup, timeout recovery), the `safeWrite` can race — another process may read the file before `dispatched_to` is set and write it back afterward, losing the field.

**Fix:** Added an atomic `mutateJsonFileLocked` write in `spawnAgent()` immediately after moving the dispatch entry from pending → active. This stamps `dispatched_to` and `dispatched_at` on the originating work item using the correct locked read-modify-write pattern. The write uses `||=` semantics so it doesn't overwrite if the fields are already set by the discover phase.

**Affected paths:**
- `engine.js:spawnAgent()` — new atomic write after pending→active transition
- Handles both project work items (`source: 'work-item'`) and central work items (`source: 'central-work-item'`/`'central-work-item-fanout'`)

**Test:** Added source-pattern test `spawnAgent atomically stamps dispatched_to on work item (#402)` verifying the presence of `mutateJsonFileLocked`, `dispatched_to`, and `dispatched_at` stamps.

PR #482, opened 2026-04-07
fix(W-mnp7e5qrk09h): defer dispatched status until spawn + _any_ routing token

### LINKED ISSUE #480: Work item shows dispatched before an agent is actually working on it
## Bug

Two related issues with work item dispatch status and fix routing:

### 1. `dispatched` status is premature

When a work item is added to the dispatch queue but no agent has yet picked it up, the item is immediately written with `status: dispatched`. This is misleading � `dispatched` implies an agent is actively working on it. If all agents are busy, the item sits in the dispatch queue with `dispatched` shown in the dashboard UI until an agent eventually becomes free.

**Expected:** Status should remain `pending` until the moment an agent process is actually spawned. Only flip to `dispatched` (or a new `queued` state) after `spawnAgent()` succeeds.

### 2. Fix routing falls through to only one fallback instead of any idle agent

Routing table: `fix ? _author_ ? dallas`. For standalone fix work items (not created from a PR with a known author), `_author_` resolves to null and the item falls through to `dallas` only. If Dallas is busy, the item sits queued even though other agents (Ripley, Lambert, Rebecca, Ralph) are idle.

**Expected:** Fix tasks should prefer the original PR author, then fall back to **any idle agent** � not a single hard-coded fallback. The routing table fallback should support `_any_` as a token meaning "first available idle agent".

## Observed

Work items W-mnp6y50gwrhy, W-mnp6y84ltgvp, W-mnp6yaghm17d were created as fix tasks with `agent: undefined`, immediately showed `status: dispatched`, and sat waiting for Dallas (fallback) even while Ripley was idle.

## Suggested Fix

1. In `discoverWork` / dispatch loop: only mutate item status to `dispatched` after `spawnAgent()` returns successfully, not when the item is first selected from the queue.
2. Add `_any_` token to routing resolver � when preferred and named fallback are both busy, iterate all configured agents and pick the first idle one.

---
_Filed via Minions dashboard_

yemi33:
## Summary
- **Bug 1**: Work items were marked `dispatched` during discovery before any agent spawned. If agents were busy, items showed `dispatched` in dashboard when no agent was running. Fixed by deferring status to `spawnAgent()` which atomically stamps `dispatched_to/dispatched_at/status`. PRD sync also moved to post-spawn in the dispatch loop.
- **Bug 2**: Fix routing fell through to `dallas` only when `_author_` was null. Added `_any_` token support to the routing resolver — picks any idle agent sorted by error rate. Updated `routing.md` to use `_any_` as fallback for fix tasks.
- Self-heal in discovery only sets `DISPATCHED` when item is in `dispatch.active` (actually spawned), not when in `dispatch.pending` (queued but not yet spawned).

## Files Changed
- `engine.js` — Removed premature DISPATCHED marking in `discoverFromWorkItems` and `discoverCentralWorkItems`. Added PRD sync after successful spawn in dispatch loop. Self-heal only promotes to DISPATCHED when in `dispatch.active`.
- `engine/routing.js` — Added `_any_` token + `pickAnyIdle()` helper for routing resolution.
- `routing.md` — Changed fix fallback from `dallas` to `_any_`, documented `_any_` token.
- `test/unit.test.js` — Updated/added tests for deferred dispatch status, `_any_` routing, and self-heal behavior.

## Test plan
- [x] All 877 unit tests pass (0 failures)
- [ ] Verify fix items route to any idle agent when dallas is busy
- [ ] Verify work items show `pending` in dashboard until agent actually spawns
- [ ] Verify PRD items sync to `dispatched` only after spawn succeeds

Closes #480

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -1547,7 +1547,7 @@ function discoverFromWorkItems(config, project) {
   const items = safeJson(projectWorkItemsPath(project)) || [];
   const cooldownMs = (src.cooldownMinutes || 0) * 60 * 1000;
   const newWork = [];
-  const prdSyncQueue = [];
+  // PRD sync for dispatched status deferred to spawnAgent success (#480)
   const skipped = { gated: 0, noAgent: 0 };
   let needsWrite = false;
   const selfHealKeys = new Set(); // Collect keys for batched self-heal (1 lock instead of N)
@@ -1594,10 +1594,11 @@ function discoverFromWorkItems(config, project) {
       needsWrite = true;
     }
     if (isAlreadyDispatched(key)) {
-      if (item.status === WI_STATUS.PENDING) { item.status = WI_STATUS.DISPATCHED; needsWrite = true; }
-      if (!item.dispatched_to) {
-        const existing = getDispatch().active?.find(d => d.meta?.dispatchKey === key);
-        if (existing?.agent) { item.dispatched_to = existing.agent; needsWrite = true; }
+      // Only self-heal to DISPATCHED if actually in dispatch.active (agent spawned) (#480)
+      const existingActive = getDispatch().active?.find(d => d.meta?.dispatchKey === key);
+      if (existingActive) {
+        if (item.status === WI_STATUS.PENDING) { item.status = WI_STATUS.DISPATCHED; needsWrite = true; }
+        if (!item.dispatched_to && existingActive.agent) { item.dispatched_to = existingActive.agent; needsWrite = true; }
       }
       if (item._pendingReason !== 'already_dispatched') { item._pendingReason = 'already_dispatched'; needsWrite = true; }
       skipped.gated++; continue;
@@ -1682,12 +1683,11 @@ function discoverFromWorkItems(config, project) {
       continue;
     }
 
-    // Mark item as dispatched BEFORE adding to newWork (prevents race on next tick)
-    item.status = WI_STATUS.DISPATCHED;
-    item.dispatched_at = ts();
-    item.dispatched_to = agentId;
+    // Don't mark dispatched here — item stays pending until spawnAgent succeeds.
+    // spawnAgent() atomically stamps dispatched_to/dispatched_at/status via locked write (#480).
+    // isAlreadyDispatched(key) prevents re-discovery on subsequent ticks.
     delete item._pendingReason;
-    prdSyncQueue.push({ id: item.id, sourcePlan: item.sourcePlan });
+    needsWrite = true;
 
     newWork.push({
       type: workType,
@@ -1729,12 +1729,9 @@ function discoverFromWorkItems(config, project) {
     }
   }
 
-  // Write back updated statuses (always, since we mark items dispatched before newWork check)
-  if (newWork.length > 0 || needsWrite) {
+  // Write back updated statuses (pendingReason clears, checkpoint counts, decompose flags, etc.)
+  if (needsWrite) {
     mutateWorkItems(projectWorkItemsPath(project), () => items);
-    if (newWork.length > 0) {
-      for (const s of prdSyncQueue) syncPrdItemStatus(s.id, 'dispatched', s.sourcePlan);
-    }
   }
 
   const skipTotal = skipped.gated + skipped.noAgent;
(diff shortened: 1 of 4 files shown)


yemi33:
## Implementation Notes

### Bug 1: Premature dispatched status (main fix)

**Root cause:** `discoverFromWorkItems()` (line 1685-1688) and `discoverCentralWorkItems()` both marked items as `DISPATCHED` with `dispatched_to`/`dispatched_at` during the discovery phase, before any agent was spawned. Items then sat in `dispatch.pending` showing "dispatched" in the dashboard when no agent was running.

**Fix approach:**
1. Removed premature status writes from both discovery functions
2. `spawnAgent()` already had an atomic locked write at line 813-822 that stamps `dispatched_to/dispatched_at/status` — this is now the single source of truth
3. Self-heal blocks (when `isAlreadyDispatched(key)` is true) now check `dispatch.active` instead of blindly promoting pending items to dispatched
4. PRD `syncPrdItemStatus()` for dispatched state moved from discovery to the dispatch loop, after successful `spawnAgent()` return

**Key invariant:** Work items stay `pending` until `spawnAgent()` succeeds. `isAlreadyDispatched()` prevents re-discovery on subsequent ticks.

### Bug 2: `_any_` routing token

**Root cause:** routing.md had `fix → _author_ → dallas`. When `_author_` was null, only dallas was tried as fallback. The existing "any idle" fallback code DID exist after both preferred/fallback failed, but making it explicit via `_any_` token gives more control.

**Fix:** Added `_any_` token support to `resolveAgent()` via `pickAnyIdle()` helper. Updated `routing.md` fallback for fix from `dallas` to `_any_`.

### Tests
- 877 tests pass, 0 failures
- Updated 3 existing tests, added 2 new tests

yemi33:
## Code Review: PR-482 — Defer dispatched status until spawn + `_any_` routing token

### Verdict: ✅ APPROVE

Solid, well-reasoned fix for two related bugs. The changes are architecturally consistent with the existing codebase patterns.

---

### Bug 1: Deferred Dispatched Status

**Problem:** `discoverFromWorkItems` and `discoverCentralWorkItems` eagerly stamped `DISPATCHED` / `dispatched_at` / `dispatched_to` during discovery, before any agent was spawned. Items in `dispatch.pending` (queued but not yet spawned) appeared as "dispatched" in the dashboard.

**Fix analysis:** The discovery phase no longer stamps `DISPATCHED`. Items stay `PENDING` until `spawnAgent()` succeeds, which then atomically stamps via `mutateJsonFileLocked` at `engine.js:812-822`. This is correct — `spawnAgent` already had this atomic stamping logic (added in #402), so removing the premature stamps eliminates the double-write and makes the status reliable.

**Self-heal guard** (`engine.js:1596-1604`, `engine.js:2020-2029`): Now correctly checks `dispatch.active` before promoting to DISPATCHED, rather than promoting when the item is merely in `dispatch.pending`. Clean refactor.

**PRD sync deferral** (`engine.js:2659-2663`): Moved from discovery to the dispatch loop post-spawn. Wrapped in try-catch. Correct.

**Spawn-failure path verified:** If spawn fails (worktree error, etc.), `spawnAgent` calls `completeDispatch(ERROR)` and returns null. The defensive re-queue at `engine.js:2643` checks `wi.status === WI_STATUS.DISPATCHED`, which won't trigger since the item was never promoted. But that's correct — the item is already `PENDING` and will be re-discovered on the next tick after cooldown. No status corruption.

### Bug 2: `_any_` Routing Token

**Implementation** (`engine/routing.js:124-143`): `pickAnyIdle(exclude)` helper is clean — filters by availability, sorts by error rate, claims the selected agent. The `_any_` resolution integrates correctly with the existing preferred→fallback→idle cascade.

**Edge cases verified:**
- `_author_` resolves to null + `_any_` fallback → `preferred = null`, falls through to `fallback === '_any_'` → `pickAnyIdle([null])` → `filter(Boolean)` strips null → correct
- Both preferred and fallback are `_any_` → first call finds nothing, second call also finds nothing (harmless `'_any_'` in exclude set, never matches a real agent ID) → correct
- `_any_` as preferred picks first idle → claimed → subsequent fallback check won't re-pick → correct

### Minor Non-Blocking Observation

**Fan-out `dispatched_to` cosmetic change:** Previously, fan-out wrote `dispatched_to: idleAgents.map(a => a.id).join(', ')` (all agent IDs comma-separated). Now, `spawnAgent` stamps `wi.dispatched_to = wi.dispatched_to || agentId`, so only the **first** agent to spawn gets recorded in `dispatched_to`. The `fanOutAgents` array still captures all agents, so no data loss — but dashboard display of fan-out items may show a single agent instead of all. Worth noting for future dashboard polish.

### Tests

Tests updated correctly. Source-pattern tests match the new code structure. The negative assertions (`!src.includes('item.dispatched_at = ts()')`) are technically whole-file scoped, but `spawnAgent` uses `wi.dispatched_at` (different variable name), so no false positives. Acceptable.

New test coverage:
- Self-heal checks `dispatch.active` (`#480`)
- Discovery doesn't prematurely stamp dispatched fields
- PRD sync deferred to dispatch loop
- `_any_` token support in routing
- `pickAnyIdle` fallback behavior

### Summary

| Area | Assessment |
|------|-----------|
| Architecture | ✅ Correct — leverages existing `spawnAgent` atomic stamping |
| Status constants | ✅ Uses `WI_STATUS.DISPATCHED` — no magic strings |
| Locking | ✅ No new lock scope issues |
| Error handling | ✅ Spawn failures handled correctly at both levels |
| Routing | ✅ `_any_` token integrates cleanly with cascade |
| Tests | ✅ 877 passing, new coverage for both bugs |
| Cross-platform | ✅ No platform-specific changes |

No blocking issues found. Ship it.

---

yemi33 (commented):
✅ **APPROVE** — Deferred dispatched status and _any_ routing token are both correctly implemented. No blocking issues. See detailed review in comment thread.

PR #806, opened 2026-04-10
fix: handle force-pushed dep branches on retry

yemi33:
Closes yemi33/minions#738

## Summary

- When `git merge origin/<dep>` fails (e.g. diverged history from a force-pushed/rebased dep branch), the engine now:
  1. Aborts the partial merge (`git merge --abort`)
  2. Resets the worktree to `origin/<mainBranch>` (`git reset --hard`)
  3. Re-merges **all** fetched dep branches from scratch on the clean base
- If the re-merge also fails (genuine conflict), marks `depMergeFailed` and bails as before (retry next tick)
- Previously, the merge failure immediately set `depMergeFailed = true`, causing a permanent retry loop when a dep was force-pushed

## Test plan

- [x] 2 new source-pattern tests verify the reset-and-re-merge code paths exist
- [x] `npm test` passes (1297 pass, 12 pre-existing failures, 0 regressions)
- [ ] Manual: create work item A depending on B, dispatch A, force-push B, retry A — should succeed

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -493,8 +493,25 @@ async function spawnAgent(dispatchItem, config) {
                 await execAsync(`git merge "origin/${depBranch}" --no-edit`, { ..._gitOpts, cwd: worktreePath });
                 log('info', `Merged dependency branch ${depBranch} (${prId}) into worktree ${branchName}`);
               } catch (mergeErr) {
-                log('warn', `Failed to merge dependency ${depBranch} into ${branchName}: ${mergeErr.message}`);
-                depMergeFailed = true;
+                // Merge failed — possibly due to diverged history from a force-pushed (rebased) dep branch.
+                // Abort partial merge, reset worktree to clean main base, and re-merge all deps from scratch.
+                log('warn', `Merge of ${depBranch} into ${branchName} failed: ${mergeErr.message} — attempting reset and re-merge of all deps`);
+                try { await execAsync(`git merge --abort`, { ..._gitOpts, cwd: worktreePath }); } catch (_) { /* no merge in progress */ }
+                const mainRef = sanitizeBranch(shared.resolveMainBranch(rootDir, project.mainBranch));
+                try {
+                  await execAsync(`git reset --hard "origin/${mainRef}"`, { ..._gitOpts, cwd: worktreePath });
+                  log('info', `Reset worktree ${branchName} to origin/${mainRef} for clean dep re-merge`);
+                  // Re-merge ALL fetched dep branches from scratch on clean base
+                  for (const { branch: reBranch, prId: rePrId } of fetched) {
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### What changed
**engine.js** (lines ~495-515): Replaced the simple `depMergeFailed = true` in the dep merge catch block with a reset-and-re-merge strategy:

1. `git merge --abort` — clean up any partial merge state
2. `git reset --hard origin/<mainBranch>` — bring worktree back to clean base
3. Re-merge ALL fetched dep branches from scratch (not just the failed one)
4. If re-merge fails too → `git merge --abort` + `depMergeFailed = true` (same bail path as before)

### Why this works
On retry, the worktree already has old dep commits merged in. When a dep is force-pushed (rebased), `git merge origin/<dep>` sees diverged history and fails. By resetting to origin/main and re-merging all deps fresh, we get the latest version of every dep branch regardless of force-pushes.

### Risk
Low. The reset-and-re-merge only triggers on merge failure (not the happy path). The fallback to `depMergeFailed` still works for genuine conflicts.

yemi33:
## Review: PR-806 — Fix #738: dep merge fails on retry when dependency branch was force-pushed

### Verdict: **APPROVE** ✅

### Summary

Clean, well-scoped fix. When a dependency merge fails (e.g., due to force-pushed/rebased dep branch with diverged history), the engine now:
1. Aborts any partial merge
2. Resets the worktree to `origin/<mainBranch>`
3. Re-merges ALL fetched dep branches from scratch on the clean base
4. Falls back to `depMergeFailed = true` if re-merge also fails (genuine conflict)

### What I checked

- **Logic flow** (`engine.js:495-516`): The recovery path correctly aborts partial merge, resets to main, and re-merges all deps (not just remaining ones). The `break` after recovery exits the original loop regardless of success/failure, which is correct — either all deps are re-merged or we bail.
- **Error handling**: Both `git merge --abort` calls are wrapped in try-catch with empty catch (appropriate — merge may not be in progress). The `resetErr` catch properly sets `depMergeFailed = true` and aborts any partial merge from the re-merge attempt.
- **`resolveMainBranch` + `sanitizeBranch`**: Correctly resolves the main branch ref rather than hardcoding `master`/`main`.
- **Tests** (`test/unit.test.js:6141-6171`): Source-pattern tests consistent with surrounding test style. Cover the abort, reset, re-merge, and failure paths.
- **No magic strings/numbers**: Uses existing constants and helpers (`DISPATCH_RESULT.ERROR`, `resolveMainBranch`, `sanitizeBranch`).

- The re-merge loop re-merges the branch that originally failed, which is intentional — the force-pushed ref was already fetched successfully, so the fresh `origin/<branch>` ref should merge cleanly onto the reset base. Correct design.

yemi33 (commented):
APPROVE — Clean fix with correct recovery logic for force-pushed dep branches. Review by Ripley.

PR #807, opened 2026-04-10
fix(W-mnt7oquq9jr7): PRD item retry when work item missing

### LINKED ISSUE #781: prd: retry/reset for PRD item fails silently when work item doesn't exist — no re-materialization path
## Bug

When a PRD item has no corresponding work item (deleted during cleanup, or orphaned), there is no way to reset or retry it from the dashboard. The retry button doesn't appear, and the engine won't re-materialize it unless the PRD item's raw status is `missing` or `planned`. Manual file editing is required.

### 1. Retry button requires work item to exist
`render-prd.js:175` — retry button visibility:
```js
wi && (wi.status === 'failed' || i.status === 'failed')
```
If `wi` is undefined (no work item), the button never renders regardless of the PRD item's status.

### 2. `/api/work-items/retry` can't create — only resets existing
`dashboard.js:1023` — `handleWorkItemsRetry` finds the work item first and errors if not found. There is no fallback to create a new work item from the PRD item definition when none exists.

### 3. `autoCleanPrdWorkItems` deletes work items without resetting PRD status
`engine.js:1160` — deletes failed/pending work items but never sets the PRD `missing_features[].status` back to `missing`. After cleanup the PRD item retains `failed`/`dispatched` status, which is not in the engine's materialization filter (`['missing', 'planned']`), so it's never re-queued.

### 4. Engine materializes only `missing`/`planned` items
`engine.js:1380`:
```js
const statusFilter = ['missing', 'planned'];
```
`failed`, `dispatched`, `pending` PRD statuses are all skipped. After a work item is deleted, the PRD item is permanently stuck unless someone manually edits the JSON.

## Affected scenario (observed today)

`P-6b1e8d4f` and `P-a6e1b5c4` were deleted during failed-item cleanup. PRD retained `dispatched` status. No retry button, no auto-re-queue. Required manual PRD JSON edits to set status to `missing` to unblock.

## Required fix

`/api/work-items/retry` (and the dashboard retry button) should handle the case where no work item exists for a given PRD item ID:

1. **Button:** show retry for PRD items where `!wi && i.status` is terminal (`failed`, `dispatched` with no active dispatch) — not just when `wi` exists
2. **Handler:** if no work item found, create a new pending work item from the PRD item definition (same logic as `materializePlansAsWorkItems` but on-demand), then proceed
3. **`autoCleanPrdWorkItems`:** when deleting a work item, also reset the PRD item status to `missing` via `syncPrdItemStatus`

Relates to #779 (orphaned dispatch status).

yemi33:
Closes yemi33/minions#781

## Summary

When a PRD item has no corresponding work item (deleted during cleanup or orphaned), there was no way to reset or retry it from the dashboard. This PR fixes all three root causes:

1. **Dashboard retry button** (`render-prd.js`): Now shows for PRD items where the work item was deleted — both list and graph views. Previously required `wi` to exist.

2. **Retry API** (`dashboard.js`): `/api/work-items/retry` now accepts a `prdFile` parameter. When no work item is found but `prdFile` is provided, it creates a new pending work item from the PRD item definition (same logic as `materializePlansAsWorkItems`), resets PRD status, and clears dispatch history/cooldowns.

3. **Auto-cleanup** (`engine.js`): `autoCleanPrdWorkItems` now calls `syncPrdItemStatus(id, 'missing', prdFile)` after deleting work items, so the engine re-materializes them on the next tick instead of leaving PRD items permanently stuck.

## Test plan

- [x] 5 new unit tests covering all three fixes (1300 passing, 12 pre-existing failures)
- [ ] Manual: delete a work item linked to a PRD item → verify retry button appears
- [ ] Manual: click retry on orphaned PRD item → verify work item is re-created
- [ ] Manual: trigger autoCleanPrdWorkItems → verify PRD status resets to 'missing'

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -1046,7 +1046,74 @@ const server = http.createServer(async (req, res) => {
           }
         }
       }
-      if (!wiPath) return jsonReply(res, 404, { error: 'work item not found in any source' });
+      // If no work item found, attempt to re-materialize from PRD item definition
+      if (!wiPath) {
+        const prdFile = body.prdFile;
+        if (!prdFile) return jsonReply(res, 404, { error: 'work item not found in any source' });
+
+        // Look up PRD item to create a new work item on-demand
+        const prdPath = path.join(PRD_DIR, prdFile);
+        const plan = shared.safeJson(prdPath);
+        if (!plan?.missing_features) return jsonReply(res, 404, { error: 'PRD file not found or invalid' });
+        const prdItem = plan.missing_features.find(f => f.id === id);
+        if (!prdItem) return jsonReply(res, 404, { error: 'PRD item not found in ' + prdFile });
+
+        // Determine target work-items file (project from PRD item or plan, fallback to central)
+        const projName = prdItem.project || plan.project || prdFile.replace(/-\d{4}-\d{2}-\d{2}\.json$/, '');
+        const proj = PROJECTS.find(p => p.name?.toLowerCase() === projName.toLowerCase());
+        const targetWiPath = proj ? shared.projectWorkItemsPath(proj) : path.join(MINIONS_DIR, 'work-items.json');
+
+        // Create new work item from PRD item definition (same logic as materializePlansAsWorkItems)
+        const complexity = prdItem.estimated_complexity || 'medium';
+        const criteria = (prdItem.acceptance_criteria || []).map(c => `- ${c}`).join('\n');
+        const newItem = {
+          id,
+          title: `Implement: ${prdItem.name}`,
+          type: complexity === 'large' ? 'implement:large' : 'implement',
+          priority: prdItem.priority || 'medium',
+          description: `${prdItem.description || ''}\n\n**Plan:** ${prdFile}\n**Plan Item:** ${prdItem.id}\n**Complexity:** ${complexity}${criteria ? '\n\n**Acceptance Criteria:**\n' + criteria : ''}`,
+          status: WI_STATUS.PENDING,
+          created: new Date().toISOString(),
+          createdBy: 'dashboard:prd-retry',
+          sourcePlan: prdFile,
+          depends_on: prdItem.depends_on || [],
+          branchStrategy: plan.branch_strategy || 'parallel',
+          featureBranch: plan.feature_branch || null,
+          project: prdItem.project || plan.project || null,
+          _source: proj?.name || 'central',
+          _retryCount: 0,
+        };
+        mutateWorkItems(targetWiPath, items => { items.push(newItem); });
+
+        // Reset PRD item status to pending
+        try {
+          const lifecycle = require('./engine/lifecycle');
+          lifecycle.syncPrdItemStatus(id, WI_STATUS.PENDING, prdFile);
+        } catch (e) { console.error('PRD status sync:', e.message); }
+
+        // Clear dispatch history and cooldowns for this item
+        const dispatchPath = path.join(MINIONS_DIR, 'engine', 'dispatch.json');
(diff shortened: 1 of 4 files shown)


yemi33:
## Implementation Notes

### Changes by file

**`engine.js` — `autoCleanPrdWorkItems`** (line ~1188)
- After deleting pending/failed work items, now calls `syncPrdItemStatus(id, 'missing', prdFile)` for each deleted ID
- This ensures PRD items don't get stuck in `failed`/`dispatched` status after cleanup

**`dashboard.js` — `handleWorkItemsRetry`** (line ~1049)
- Added re-materialization path: when no work item found and `prdFile` is provided, looks up the PRD item definition and creates a new pending work item on-demand
- Uses same work item creation logic as `materializePlansAsWorkItems` (complexity, acceptance criteria, branch strategy, etc.)
- Also clears dispatch history and cooldowns for the item
- Returns `{ ok: true, id, rematerialized: true }` so the frontend can distinguish

**`dashboard/js/render-prd.js`**
- List view `canRequeue`: now also true when `!wi && i.status` is terminal (not missing/done/planned)
- Graph view: same logic applied to the IIFE that renders retry buttons
- `prdItemRequeue()`: now accepts third `prdFile` parameter, passes it to `/api/work-items/retry`
- Both views pass `i.source` (the PRD filename) as `prdFile` when no work item exists

**`test/unit.test.js`** — 5 new tests:
- PRD list view retry for orphaned items
- PRD graph view retry for orphaned items
- `prdItemRequeue` sends `prdFile` parameter
- `handleWorkItemsRetry` re-materialization path
- `autoCleanPrdWorkItems` resets PRD status to missing

yemi33:
## Review: PR-807 — Fix #781: PRD retry/reset for missing work items

**Verdict: APPROVE** ✅

### Summary
Solid fix for a real gap: when a work item gets deleted but its PRD item remains, the retry button was silently failing (404). This PR adds re-materialization from PRD item definitions and surfaces the retry button for orphaned PRD items.

### What's Good
- **Re-materialization logic** mirrors `materializePlansAsWorkItems` patterns correctly — tagged with `createdBy: 'dashboard:prd-retry'` for traceability
- **Both views updated** (list + graph) with consistent `canRequeue`/`canRetry` logic
- **`autoCleanPrdWorkItems` now resets PRD status to `'missing'`** so the engine can re-materialize on next tick — good defensive addition
- **Tests cover all new paths** (orphaned items, prdFile param, re-materialization, PRD status reset)
- Uses `mutateWorkItems` correctly for the new work item write

### Minor Non-Blocking Notes

1. **Cooldown cleanup uses `safeWrite` instead of `mutateJsonFileLocked`** (dashboard.js ~L1107-1112). Per CLAUDE.md conventions, shared JSON files should use locked writes. In practice, cooldowns.json has low contention so this is unlikely to cause issues, but it's technically a race condition. Could be addressed in a follow-up.

2. **`syncPrdItemStatus` called with string literal `'missing'`** in `autoCleanPrdWorkItems` (engine.js ~L1188). This is valid since `syncPrdItemStatus` accepts `'missing'` as a special case beyond `WI_STATUS`, but worth noting it's one of the few places a raw string is used intentionally.

3. **Project name fallback** in the re-materialization path (`prdFile.replace(/-\d{4}-\d{2}-\d{2}\.json$/, '')`) is a reasonable heuristic but could mismatch if PRD filenames don't follow this convention. Low risk given current naming patterns.

---

yemi33 (commented):
APPROVE — Clean fix with good test coverage. Minor non-blocking suggestions noted in review comment.

PR #821, opened 2026-04-11
fix(W-mntkzthce4r2): PID file leak in indirect spawn + kb-sweep race condition

yemi33:
## Summary

- **PID file leak fix (engine/llm.js)**: The indirect spawn path creates prompt/sys temp files but `spawn-agent.js` also derives a PID file from the prompt path. This PID file was never added to `cleanupFiles`, causing hundreds of leaked `*-pid-*.pid` files in `engine/tmp/`. Fixed by deriving the same PID path and including it in cleanup.
- **kb-sweep race condition fix (dashboard.js)**: `handleKnowledgeSweep` auto-releases `_kbSweepInFlight` after 5 minutes, but the original sweep's `finally` block still fires later and clears the flag for any newly-started sweep, allowing unbounded concurrent sweeps. Fixed with a generation token (`sweepToken`) so `finally` only clears the flag if this sweep is still the current owner.
- **Cleanup**: Removed 484 leaked PID files from `engine/tmp/`
- **Tests**: Added 2 source-pattern tests verifying both fixes

## Test plan
- [x] `npm test` — 1329 passed, 0 failed, 2 skipped
- [ ] Verify no new PID file leaks after running kb-sweep or command-center calls
- [ ] Verify concurrent kb-sweep requests are properly guarded (second request gets 409)

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -1876,6 +1876,9 @@ const server = http.createServer(async (req, res) => {
       global._kbSweepInFlight = false;
     }
     if (global._kbSweepInFlight) return jsonReply(res, 409, { error: 'sweep already in progress' });
+    // Generation token prevents stale finally blocks from clearing the flag for a new sweep
+    const sweepToken = Date.now() + Math.random();
+    global._kbSweepToken = sweepToken;
     global._kbSweepInFlight = true;
     global._kbSweepStartedAt = Date.now();
     try {
@@ -2017,7 +2020,7 @@ If nothing to do: { "duplicates": [], "reclassify": [], "remove": [] }`;
       const summary = `${merged} duplicates merged, ${removed} stale removed, ${reclassified} reclassified${pruned ? ', ' + pruned + ' old swept files pruned' : ''}`;
       safeWrite(path.join(ENGINE_DIR, 'kb-swept.json'), JSON.stringify({ timestamp: new Date().toISOString(), summary }));
       return jsonReply(res, 200, { ok: true, summary, plan });
-    } catch (e) { return jsonReply(res, 500, { error: e.message }); } finally { global._kbSweepInFlight = false; }
+    } catch (e) { return jsonReply(res, 500, { error: e.message }); } finally { if (global._kbSweepToken === sweepToken) global._kbSweepInFlight = false; }
   }
 
   async function handlePlansList(req, res) {
--- a/engine/llm.js
+++ b/engine/llm.js
@@ -108,7 +108,9 @@ function _spawnProcess(promptText, sysPromptText, { direct, label, model, maxTur
   const sysPath = path.join(tmpDir, `${label}-sys-${id}.md`);
   safeWrite(promptPath, promptText);
   safeWrite(sysPath, sysPromptText || '');
-  cleanupFiles.push(promptPath, sysPath);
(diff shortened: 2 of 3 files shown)


yemi33:
## Implementation Notes

### Bug 1 — PID file leak (engine/llm.js:107-113)
The indirect spawn path in `_spawnProcess()` creates `{label}-prompt-{id}.md` and `{label}-sys-{id}.md` temp files. However, `spawn-agent.js:130` derives a PID file: `promptFile.replace(/prompt-/, 'pid-').replace(/\.md$/, '.pid')` → `{label}-pid-{id}.pid`. This PID file was never included in `cleanupFiles`, so on process close it was never cleaned up.

**Fix**: Derive the same PID path inline and add it to `cleanupFiles`:
```js
const pidPath = promptPath.replace(/prompt-/, 'pid-').replace(/\.md$/, '.pid');
cleanupFiles.push(promptPath, sysPath, pidPath);
```

**Note**: The kb-sweep calls use `direct: true` so they don't go through spawn-agent.js — meaning they don't create PID files. The 484 leaked files were actually from `command-center` (357), `doc-chat`, and other labels using the indirect path.

### Bug 2 — _kbSweepInFlight race condition (dashboard.js:1879-2023)
The auto-release after 5 minutes clears `_kbSweepInFlight`, allowing a new sweep to start. But the **original sweep's** `finally { global._kbSweepInFlight = false; }` still fires when it eventually completes, clearing the flag mid-sweep for the new one.

**Fix**: Generation token pattern — each sweep gets a unique `sweepToken`. The `finally` block only clears the flag if `global._kbSweepToken === sweepToken` (i.e., this sweep is still the current owner).

### Cleanup
Removed 484 leaked `*-pid-*.pid` files from `engine/tmp/` on the main working tree.

yemi33:
## Review: PR-821 — Fix kb-sweep: PID file leak + _kbSweepInFlight race condition

### Verdict: ✅ APPROVE

### Summary

Two clean, targeted fixes:

**1. PID file leak in `engine/llm.js`**
The indirect spawn path (via `spawn-agent.js`) creates a PID file derived from the prompt path, but the caller never cleaned it up. Fix correctly mirrors the derivation logic from `spawn-agent.js:130` (`prompt-` → `pid-`, `.md` → `.pid`) and adds it to `cleanupFiles`.

**2. `_kbSweepInFlight` race condition in `dashboard.js`**
A stale `finally` block from a previous sweep could clear the flag for a *new* sweep that started after the 409 guard but before the old `finally` ran. The generation token (`sweepToken`) ensures only the owning sweep clears the flag. Simple and correct.

### Code Quality
- Both fixes follow existing patterns — no new abstractions, no unnecessary complexity
- PID path derivation is an exact copy of `spawn-agent.js:130` (verified)
- Tests are source-pattern style (consistent with surrounding tests in this section)
- No magic strings, no lock issues, no cross-platform concerns

### Minor Notes (non-blocking)
- The `sweepToken` uses `Date.now() + Math.random()` which is fine for uniqueness here, though a simple incrementing counter would also work. Not worth changing.

No issues found. Ship it.

---

yemi33 (commented):
APPROVE — Both fixes are correct, minimal, and well-tested. PID cleanup matches spawn-agent.js derivation exactly. Sweep token prevents stale finally race. Ship it. (Note: GitHub blocked formal approval on own PR)

PR #822, opened 2026-04-11
fix: add checkTimeouts ReferenceError smoke test (closes #775)

### LINKED ISSUE #775: engine: checkTimeouts crashes with 'DEFAULT_HEARTBEAT_TIMEOUTS is not defined' — hung agent detection blind for entire session
## Bug

`checkTimeouts` throws `ReferenceError: DEFAULT_HEARTBEAT_TIMEOUTS is not defined` on every tick. The `safe()` wrapper catches it and logs the warning, but this silently disables ALL hung-agent detection for the entire engine session.

## Impact

- Agents that hang (no output) are never killed or retried
- Work items remain stuck in `dispatched` state indefinitely
- Max-turns limit becomes the only backstop (50 turns = hours of wasted compute and cost)

## Observed

685 occurrences across 3 engine restarts between `2026-04-09T19:17` and `2026-04-09T21:18`. Every 30s tick logged:
```
warn checkTimeouts: DEFAULT_HEARTBEAT_TIMEOUTS is not defined
```

During this window, agents ran to `error_max_turns` (50 turns, ~$5+ each) instead of being killed after the `heartbeatTimeout` (5 min default).

## Root cause

`checkTimeouts` in `engine/timeout.js` references a constant `DEFAULT_HEARTBEAT_TIMEOUTS` that doesn't exist in the module scope. The constant was removed/renamed at some point but the reference wasn't cleaned up. The CHANGELOG (`0.1.x`) mentions "remove undefined DEFAULT_HEARTBEAT_TIMEOUTS" — this fix wasn't applied to the running session until a later restart.

## Fix

Find and remove/replace the stale `DEFAULT_HEARTBEAT_TIMEOUTS` reference in `engine/timeout.js`. Add a smoke test that calls `checkTimeouts({})` without a full engine context and asserts no ReferenceError — this would have caught the regression.

## Affected session

Engine sessions `2026-04-09T19:17` → `2026-04-09T21:18`. New session (`2026-04-10T06:46`) does not reproduce (fix was already in deployed code).

yemi33:
Closes yemi33/minions#775

## Summary
- The code fix for `DEFAULT_HEARTBEAT_TIMEOUTS` was already applied in 9bbabd6, but existing regression tests only did source-string matching (`src.includes(...)`)
- Adds a **behavioral smoke test** that actually calls `checkTimeouts({})` and asserts no `ReferenceError` is thrown — this catches stale/undefined constant references at runtime, not just in source text
- The test allows `TypeError` and other expected errors from missing engine context, but specifically fails on `ReferenceError` (the class of bug from #775)

## Test plan
- [x] New smoke test passes: `checkTimeouts({}) does not throw ReferenceError (#775 smoke test)`
- [x] Full test suite: 1296 passed, 12 failed (all pre-existing), 2 skipped

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -4059,6 +4059,22 @@ async function testCheckTimeouts() {
     assert.ok(perTypeTimeouts !== null, 'perTypeTimeouts should be constructable without ReferenceError');
     assert.strictEqual(perTypeTimeouts.review, 600000, 'config overrides should merge correctly');
   });
+
+  // Smoke test: actually call checkTimeouts({}) to catch ReferenceErrors at runtime (#775)
+  // Previous tests only checked source strings — this exercises the real function.
+  await test('checkTimeouts({}) does not throw ReferenceError (#775 smoke test)', () => {
+    const timeout = require('../engine/timeout');
+    try {
+      timeout.checkTimeouts({});
+    } catch (e) {
+      // TypeError, missing engine context, etc. are expected in a test harness.
+      // ReferenceError means a stale/undefined constant reference — that's the bug.
+      if (e instanceof ReferenceError) {
+        throw new Error(`checkTimeouts threw ReferenceError (stale constant?): ${e.message}`);
+      }
+      // Any other error is fine — the function needs a full engine context to run end-to-end.
+    }
+  });
 }
 
 // ─── engine.js — addToDispatch Tests ────────────────────────────────────────


yemi33:
## Implementation Notes

**Root cause:** The `DEFAULT_HEARTBEAT_TIMEOUTS` constant was removed/renamed but a stale reference in `checkTimeouts` wasn't cleaned up. The `safe()` wrapper in `engine.js` caught the `ReferenceError` silently, disabling ALL hung-agent detection for the entire session. This was already fixed in commit 9bbabd6.

**What this PR adds:** A behavioral smoke test that actually calls `checkTimeouts({})` and asserts no `ReferenceError`. The existing regression tests (#721) only did source-string matching (`src.includes('DEFAULT_HEARTBEAT_TIMEOUTS')`), which wouldn't catch a new undefined constant introduced in the future. The smoke test exercises the real function at runtime.

**Test design:** `checkTimeouts` needs a full engine context (activeProcesses, dispatch, etc.) to run end-to-end, so the test catches all errors but only fails on `ReferenceError` — the specific class of bug that caused #775. Other errors (TypeError from missing context) are expected and allowed.

yemi33:
## Review: PR-822 — Fix #775: checkTimeouts ReferenceError regression test

**Verdict: APPROVE** ✅

### Summary

The actual code fix for `DEFAULT_HEARTBEAT_TIMEOUTS` was already applied in `9bbabd6`. This PR adds a behavioral smoke test that calls `checkTimeouts({})` and asserts no `ReferenceError` is thrown — catching stale/undefined constant references at runtime.

### Findings

- **Test is well-designed**: Catches `ReferenceError` specifically (the bug class) while allowing other errors (expected since `checkTimeouts` needs a full engine context). Good pattern.
- **Follows existing conventions**: Placed correctly inside `testCheckTimeouts()`, uses the project's `await test()` pattern.
- **No issues found**: Clean, minimal change. No magic strings, no cross-platform concerns.

### Minor Note (non-blocking)

The commit message clearly documents the relationship to the prior fix commit (`9bbabd6`), which is good practice for two-part fixes.

---

PR #886, opened 2026-04-11
fix(timeout): short-circuit restart grace period for verifiably-dead agents (closes #869)

### LINKED ISSUE #869: fix(timeout): short-circuit restart grace period for verifiably-dead agent processes
## Bug

When the engine restarts and finds an active dispatch entry whose process is no longer alive, it correctly marks the agent as *unattached* and starts a 20-minute restart grace period (`restartGracePeriod: 1200000`). During this window, the orphan-detection guard in `timeout.js:309` is suppressed:

\`\`\`js
if (!hasProcess && silentMs > effectiveTimeout && Date.now() > engineRestartGraceUntil)
\`\`\`

This means a verifiably-dead agent (no PID in tasklist, no live-output activity) can show as **"working"** in the dashboard for up to 20 minutes after the engine restarts.

## Reproduction

1. Spawn an agent.
2. Kill the agent process externally (or let it crash).
3. Wait > 5 minutes (so live-output log is stale beyond the 5-min re-attach window).
4. Restart the engine.
5. Dashboard shows the agent as **working** for up to 20 minutes.

## Root Cause

The grace period is intentionally global — it protects legitimately-running agents whose PID files haven't been written yet or whose live-output is momentarily quiet. But agents with a *known-dead* PID (verified via \`tasklist\` on Windows / \`process.kill(pid, 0)\` on Unix) should be excluded from the grace period entirely.

The PID check already exists at restart time (\`cli.js:166-172\`) but its result is not propagated to \`checkTimeouts\`. Once the engine is running, \`checkTimeouts\` only looks at \`activeProcesses.has(id)\` — not whether the PID itself is alive.

## Fix

In \`engine/timeout.js\`, when \`!hasProcess\` and the grace period is still active, perform a fast PID liveness check using the PID stored in the dispatch item or its PID file. If the PID is confirmed dead, bypass \`engineRestartGraceUntil\` and proceed with orphan handling immediately.

Alternatively, during the restart re-attach loop in \`cli.js\`, record which dispatch IDs have confirmed-dead PIDs and set their \`engineRestartGraceUntil\` exemption so \`checkTimeouts\` can act on the first tick.

## Impact

- Dead agents show as "working" in the dashboard for ~20 minutes post-restart
- The plan-to-prd (or other) task is not re-dispatched until the grace period expires, blocking downstream work

yemi33:
## Summary
- When the engine restarts and finds an active dispatch with a PID confirmed dead (via `tasklist` on Windows / `process.kill(pid, 0)` on Unix), those dispatch IDs are now recorded in `engineRestartGraceExempt` Set
- `checkTimeouts` in `timeout.js` checks this exempt set and bypasses the 20-minute restart grace period for confirmed-dead agents, allowing orphan handling to fire immediately
- Previously, dead agents would show as "working" in the dashboard for up to 20 minutes post-restart

## Changes
- **engine.js**: Added `engineRestartGraceExempt` Set and exported it
- **engine/cli.js**: In the restart re-attach loop, when a PID is found but confirmed dead via liveness check, the dispatch ID is added to `engineRestartGraceExempt`
- **engine/timeout.js**: Orphan detection condition now checks `engineRestartGraceExempt.has(item.id)` as an alternative to the global grace period expiry
- **test/unit.test.js**: 4 new source-pattern tests verifying the exempt set exists, is exported, is populated, and is checked

## Test plan
- [x] `npm test` — 1460 passed, 0 failed, 2 skipped
- [ ] Manual: kill an agent process, restart engine, verify dead agent is detected on first tick (not after 20 min)

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -146,6 +146,7 @@ const activeProcesses = new Map(); // dispatchId → { proc, agentId, startedAt 
 const realActivityMap = new Map(); // dispatchId → timestamp of last REAL agent output (not engine heartbeat)
 // tempAgents imported from engine/routing.js
 let engineRestartGraceUntil = 0; // timestamp — suppress orphan detection until this time
+const engineRestartGraceExempt = new Set(); // dispatch IDs with confirmed-dead PIDs at restart — bypass grace period
 
 // Per-tick cache of refs that failed to fetch — avoids repeating 30s ETIMEDOUT for same missing ref
 // Cleared at the start of each tick cycle (see tickInner)
@@ -3155,7 +3156,8 @@ module.exports = {
 
   // Dispatch management (re-exported from engine/dispatch.js)
   mutateDispatch, addToDispatch, isRetryableFailureReason, completeDispatch, writeInboxAlert, updateAgentStatus,
-  activeProcesses, realActivityMap, get engineRestartGraceUntil() { return engineRestartGraceUntil; },
+  activeProcesses, realActivityMap, engineRestartGraceExempt,
+  get engineRestartGraceUntil() { return engineRestartGraceUntil; },
   set engineRestartGraceUntil(v) { engineRestartGraceUntil = v; },
 
   // Agent lifecycle
--- a/engine/cli.js
+++ b/engine/cli.js
@@ -161,6 +161,7 @@ const commands = {
           }
         }
 
+        const hadPid = agentPid && agentPid > 0; // track before liveness check
         if (agentPid && agentPid > 0) {
           try {
             if (process.platform === 'win32') {
@@ -170,6 +171,11 @@ const commands = {
               process.kill(agentPid, 0);
             }
           } catch { agentPid = null; }
+        }
+
+        // PID was found but confirmed dead — exempt from restart grace period (#869)
+        if (hadPid && !agentPid) {
+          e.engineRestartGraceExempt.add(item.id);
         }
 
         if (agentPid) {
--- a/engine/timeout.js
+++ b/engine/timeout.js
@@ -127,6 +127,7 @@ function checkSteering(config) {
 function checkTimeouts(config) {
   const activeProcesses = engine().activeProcesses;
   const engineRestartGraceUntil = engine().engineRestartGraceUntil;
+  const engineRestartGraceExempt = engine().engineRestartGraceExempt;
   const { completeDispatch } = dispatch();
   const { runPostCompletionHooks } = require('./lifecycle');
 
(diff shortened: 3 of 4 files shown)


yemi33:
## Implementation Notes

### Root Cause
The `checkTimeouts` function in `timeout.js` suppresses orphan detection during the restart grace period (`engineRestartGraceUntil`). This is intentional for agents whose PIDs are unknown (no PID file, possibly still alive). But agents with a **verified-dead PID** were also suppressed — they'd show as "working" for up to 20 minutes.

### Fix Design
Minimal, surgical change across 3 files:

1. **`engine.js:149`** — New `engineRestartGraceExempt = new Set()` alongside existing `engineRestartGraceUntil`
2. **`engine/cli.js:167-169`** — In the re-attach loop, tracks `hadPid` before the liveness check. If the PID was found but confirmed dead (tasklist/kill(0) failed), adds the dispatch ID to the exempt set.
3. **`engine/timeout.js:311`** — The orphan detection condition changes from:
   ```js
   Date.now() > engineRestartGraceUntil
   ```
   to:
   ```js
   Date.now() > engineRestartGraceUntil || engineRestartGraceExempt?.has(item.id)
   ```

### Why not do the PID check in `checkTimeouts` directly?
The issue suggested two approaches. I chose the Set-at-startup approach because:
- The PID liveness check already happens at startup in `cli.js` — no need to duplicate it
- Avoids per-tick `tasklist` calls for every active dispatch item (expensive on Windows)
- Cleaner separation: cli.js records facts at startup, timeout.js acts on them

### Edge Cases
- Items handled by `detectOrphanCompletions` (already completed via output) are removed from dispatch before `checkTimeouts` runs — exempt entries for them are harmless no-ops
- Items with no PID file at all (unknown process state) are NOT exempt — the grace period correctly protects them
- The `?.` optional chaining on `engineRestartGraceExempt?.has()` guards against test environments where the engine module may not be fully initialized

yemi33:
## Review: PR-886 — Short-circuit restart grace period for verifiably-dead agents

**Verdict: ✅ APPROVE**

### Summary

Clean, well-scoped fix for #869. When the engine restarts and re-attaches to active dispatches, agents with PIDs that are confirmed dead (PID file exists but process doesn't) are now exempt from the 20-minute grace period, allowing immediate orphan detection and retry.

### What Changed

| File | Change |
|------|--------|
| `engine.js` | New `engineRestartGraceExempt` Set, exported |
| `engine/cli.js` | Re-attach loop tracks `hadPid` before liveness check; adds confirmed-dead dispatch IDs to exempt set |
| `engine/timeout.js` | Orphan detection condition now short-circuits grace period for exempt IDs |
| `test/unit.test.js` | 4 source-pattern tests covering all three files |

### Analysis

- **Logic is correct**: `hadPid` captures `agentPid > 0` *before* the liveness check. If the check nulls `agentPid`, the ID is added to the exempt set. The timeout condition `(Date.now() > engineRestartGraceUntil || engineRestartGraceExempt?.has(item.id))` correctly allows orphan detection for these specific IDs while preserving grace period for others.
- **Optional chaining** (`engineRestartGraceExempt?.has`) is a good defensive touch.
- **No lock concerns** — the Set is only written during startup re-attach (single-threaded) and read during tick (same thread).
- **Tests** are source-pattern style, consistent with surrounding tests in the same section.

### Minor Non-Blocking Note

- The `engineRestartGraceExempt` Set grows but is never cleaned up (items stay after dispatch completes). This is harmless since dispatch IDs are unique and the set is small, but a cleanup in `completeDispatch` could be added later if desired.

PR #898, opened 2026-04-11
fix(lifecycle): verify PRD file exists before marking plan-to-prd work item as done (closes #893)

### LINKED ISSUE #893: fix(lifecycle): verify PRD file exists before marking plan-to-prd work item as done
## Bug

A `plan-to-prd` work item can be marked `done` even when the target PRD file was never written. This happens when the agent exits with `code=0` after producing no meaningful output (e.g. a confused `--resume` session that outputs one line and exits).

## Example

Work item `W-mnumw4tuwhrp` (`_prdFilename: officeagent-2026-04-11-2.json`):
- Dispatched at 18:08:32 UTC
- Completed at 18:08:57 UTC (25 seconds)
- Agent output: one line — `"I'll start by reading the pinned context..."`
- Exit code: 0 → marked `done`
- Actual PRD file: **never written**

The work item is now `done` with no deliverable. The engine won't retry it.

## Root Cause

`engine/lifecycle.js` (or the `onAgentClose` handler) marks the work item `done` based solely on exit code. For `plan-to-prd` tasks, it does not verify that `_prdFilename` actually exists on disk after the agent exits.

## Fix

In the post-completion handler for `plan-to-prd` type items, after `code=0`:

1. Check `fs.existsSync(path.join(PRD_DIR, item._prdFilename))`
2. If the file is missing → treat as failure, reset work item to `pending` with a retry reason like `"PRD file not written"`, and log a warning
3. If the file exists → proceed with `done` as today

This makes success verification concrete rather than trusting exit code alone.

yemi33:
## Summary

- **Move PRD file existence check before done status assignment** using `skipDoneStatus` flag, preventing a timing window where `checkPlanCompletion` could fire on a falsely-done plan-to-prd work item
- Plan-to-prd items that exit with code=0 but produce no PRD file are now auto-retried (up to `ENGINE_DEFAULTS.maxRetries`) or marked failed — never left as `done` without a deliverable
- Remove duplicate post-hoc check that previously tried to revert done status after the fact (now redundant)
- Add unit test verifying the `skipDoneStatus` guard runs before `updateWorkItemStatus(DONE)`

## Test plan

- [x] All 1463 unit tests pass (0 failures)
- [ ] Verify plan-to-prd with confused agent (short output, code=0) is retried, not marked done
- [ ] Verify plan-to-prd with valid PRD file created is still marked done correctly

### DIFF
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -1494,6 +1494,50 @@ async function runPostCompletionHooks(dispatchItem, agentId, code, stdout, confi
     // If decomposition produced nothing, fall through to mark parent as done
   }
 
+  // Verify plan-to-prd tasks actually created their PRD file before marking done (#893)
+  if (type === WORK_TYPE.PLAN_TO_PRD && effectiveSuccess && meta?.item?.id) {
+    let prdFound = false;
+    const expectedFile = meta.item._prdFilename;
+    if (expectedFile) {
+      prdFound = fs.existsSync(path.join(PRD_DIR, expectedFile));
+    }
+    if (!prdFound && meta?.item?.planFile) {
+      // Fall back to scanning PRD_DIR for matching source_plan
+      try {
+        for (const f of fs.readdirSync(PRD_DIR)) {
+          if (!f.endsWith('.json')) continue;
+          try {
+            const prd = safeJson(path.join(PRD_DIR, f));
+            if (prd && prd.source_plan === meta.item.planFile) { prdFound = true; break; }
+          } catch {}
+        }
+      } catch {}
+    }
+    if (!prdFound) {
+      skipDoneStatus = true;
+      const wiPath = resolveWorkItemPath(meta);
+      if (wiPath) {
+        mutateJsonFileLocked(wiPath, data => {
+          if (!Array.isArray(data)) return data;
+          const w = data.find(i => i.id === meta.item.id);
+          if (!w) return data;
+          const retries = w._retryCount || 0;
+          if (retries < ENGINE_DEFAULTS.maxRetries) {
+            w.status = WI_STATUS.PENDING;
+            w._retryCount = retries + 1;
+            delete w.dispatched_at;
+            log('warn', `plan-to-prd ${meta.item.id} completed without PRD file — auto-retry ${retries + 1}/${ENGINE_DEFAULTS.maxRetries}`);
+          } else {
+            w.status = WI_STATUS.FAILED;
+            w.failReason = 'PRD file not written after ' + ENGINE_DEFAULTS.maxRetries + ' attempts';
+            log('warn', `plan-to-prd ${meta.item.id} failed — PRD file not written after ${ENGINE_DEFAULTS.maxRetries} retries`);
+          }
+          return data;
+        });
+      }
+    }
+  }
+
   if (effectiveSuccess && meta?.item?.id && !skipDoneStatus) {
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### Root Cause
The `runPostCompletionHooks` function had a **timing bug**: it marked plan-to-prd work items as `done` at line 1499 via `updateWorkItemStatus()`, then ran `checkPlanCompletion()` at line 1525 (which could trigger downstream plan completion effects), and only *afterwards* (line 1620) checked if the PRD file actually existed on disk. If the file was missing, it tried to revert — but by then, downstream effects had already fired.

### Fix
Moved the PRD file existence check **before** the `done` status assignment by using the `skipDoneStatus` mechanism (same pattern as `DECOMPOSE` type). The check:
1. Looks for `_prdFilename` on disk (deterministic)
2. Falls back to scanning `PRD_DIR` for matching `source_plan`
3. If not found: sets `skipDoneStatus = true`, resets to `pending` with retry (or `failed` after `maxRetries`)

Removed the old duplicate post-hoc check that was at lines 1620-1661 (now fully superseded).

### Files Changed
- `engine/lifecycle.js` — core fix (moved PRD check before done marking, removed duplicate)
- `test/unit.test.js` — structural test verifying `skipDoneStatus` is set before `updateWorkItemStatus(DONE)`

yemi33:
## Review: APPROVE ✅

**PR-898**: fix(lifecycle): verify PRD file exists before marking plan-to-prd work item as done

### Summary

This PR fixes #893 by moving the PRD file existence check **before** `updateWorkItemStatus(meta, WI_STATUS.DONE)`, using the existing `skipDoneStatus` flag to prevent premature done-marking when the agent completes without actually writing a PRD file.

### What changed

- **engine/lifecycle.js**: PRD check moved from after-done (useless) to before-done, setting `skipDoneStatus = true` when no PRD file found. Retry/fail logic preserved identically.
- **test/unit.test.js**: Source-pattern test verifying the check appears before the done call.

### Assessment

- ✅ Follows existing patterns (`skipDoneStatus`, `mutateJsonFileLocked`, `ENGINE_DEFAULTS.maxRetries`)
- ✅ Uses constants — no magic strings or numbers
- ✅ Fallback scan by `source_plan` preserved
- ✅ `_prdFilename` deterministic check first, scan second — correct ordering
- ✅ Test covers the ordering constraint

### Minor note (non-blocking)

`mutateJsonFileLocked` is not `await`ed, but this matches the existing pattern throughout lifecycle.js. Fire-and-forget here is acceptable since `skipDoneStatus` is set synchronously before it.

---

yemi33:
## Fix Review Response

**Review findings addressed:** The comment triggering this fix dispatch was the author's own implementation notes, not actionable review feedback. Dallas's review (APPROVE ✅) had one non-blocking note:

> `mutateJsonFileLocked` is not `await`ed

**Response:** This is intentional and matches the existing pattern throughout `lifecycle.js` (e.g., lines 1581, 1642 in the original code). `mutateJsonFileLocked` is synchronous (uses `fs.openSync` with exclusive flag for file locking). The `skipDoneStatus = true` at line 1517 is set synchronously before the lock callback, so the done-marking at line 1541 correctly sees the flag regardless.

**Build & test status:** 1463 passed, 0 failed, 2 skipped

No code changes needed — PR is clean as approved.

yemi33:
Superseded by #911

PR #911, opened 2026-04-11
fix(lifecycle): verify PRD file exists before marking plan-to-prd done (#893)

yemi33:
## Summary

- **Bug**: `plan-to-prd` work items were marked `done` even when the PRD file was never written. The existing PRD check ran *after* `updateWorkItemStatus(DONE)`, which deleted `_retryCount` — so retries never advanced past 1, and the item oscillated between `done` and `pending` indefinitely without ever reaching `maxRetries` to be marked `failed`.
- **Fix**: Moved the PRD file existence check *before* `updateWorkItemStatus(DONE)`, using the `skipDoneStatus` pattern (same as decompose). This preserves `_retryCount` across retry cycles.
- **Improvements**: Clears `completedAt` on retry (prevents `isItemCompleted()` false positive), sets `failedAt` on final failure, wraps in try-catch for resilience.

## Test plan

- [x] 5 source-pattern tests verify the check runs before done-marking, uses `skipDoneStatus`, clears `completedAt`, and old check is removed
- [x] 5 behavioral tests with isolated `createTestMinionsDir()`: missing PRD → pending, existing PRD → done, retry count advances, maxRetries → failed, source_plan fallback scan
- [x] All 1474 tests pass (10 new, 0 failures)

### DIFF
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -1499,6 +1499,54 @@ async function runPostCompletionHooks(dispatchItem, agentId, code, stdout, confi
     // If decomposition produced nothing, fall through to mark parent as done
   }
 
+  // Verify plan-to-prd actually created the PRD file before marking done (#893)
+  // Must run BEFORE updateWorkItemStatus(DONE) — otherwise _retryCount is deleted and retries never advance
+  if (effectiveSuccess && type === WORK_TYPE.PLAN_TO_PRD && meta?.item?.id) {
+    let prdFound = false;
+    const expectedFile = meta.item._prdFilename;
+    if (expectedFile) {
+      prdFound = fs.existsSync(path.join(PRD_DIR, expectedFile));
+    }
+    if (!prdFound && meta.item.planFile) {
+      try {
+        for (const f of fs.readdirSync(PRD_DIR)) {
+          if (!f.endsWith('.json')) continue;
+          try {
+            const prd = safeJson(path.join(PRD_DIR, f));
+            if (prd && prd.source_plan === meta.item.planFile) { prdFound = true; break; }
+          } catch {}
+        }
+      } catch {}
+    }
+    if (!prdFound) {
+      skipDoneStatus = true;
+      const wiPath = resolveWorkItemPath(meta);
+      if (wiPath) {
+        try {
+          mutateJsonFileLocked(wiPath, data => {
+            if (!Array.isArray(data)) return data;
+            const w = data.find(i => i.id === meta.item.id);
+            if (!w) return data;
+            const retries = w._retryCount || 0;
+            if (retries < ENGINE_DEFAULTS.maxRetries) {
+              w.status = WI_STATUS.PENDING;
+              w._retryCount = retries + 1;
+              delete w.dispatched_at;
+              delete w.completedAt;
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### Root Cause

The existing PRD file check at lines 1625-1666 of `lifecycle.js` ran **after** `updateWorkItemStatus(meta, WI_STATUS.DONE, '')` (line 1504). The `updateWorkItemStatus` function deletes `_retryCount` when setting status to `done` (line 603). The subsequent PRD check reads `_retryCount || 0` and always finds 0, so the retry counter never advances past 1. The item oscillates between `done` → `pending (retry 1)` → `done` → `pending (retry 1)` infinitely.

### Fix

Moved the PRD check to **before** the done-marking, using `skipDoneStatus = true` (same pattern as decompose handler at line 1496). This means:

1. `updateWorkItemStatus(DONE)` never runs for plan-to-prd items without a PRD file
2. `_retryCount` is preserved on the work item across retry cycles
3. After `maxRetries` (3) failed attempts, the item correctly transitions to `failed`

### Additional Hardening

- `delete w.completedAt` on retry — prevents `isItemCompleted()` from returning `true` for a `pending` item
- `w.failedAt = ts()` on final failure — consistent with other failure paths
- `try-catch` around the `mutateJsonFileLocked` call — prevents post-completion hook crash if file lock fails
- Removed the old duplicate check (was dead code after the move)

### Test Coverage

10 new tests: 5 source-pattern + 5 behavioral with full isolation via `createTestMinionsDir()`.

yemi33:
## Review: PR-911 — fix(lifecycle): verify PRD file exists before marking plan-to-prd done (#893)

**Verdict: APPROVE** ✅

### Summary

This PR fixes a timing bug where the plan-to-prd PRD file existence check ran *after* `updateWorkItemStatus(DONE)`, which deleted `_retryCount` — making retries impossible. The fix moves the check before the done-marking call and uses `skipDoneStatus` to gate it.

### What Changed

1. **Moved PRD check before `updateWorkItemStatus(DONE)`** — the core fix. Now `_retryCount` is preserved when the check reads it.
2. **`skipDoneStatus = true`** prevents the item from being marked done when PRD is missing.
3. **`delete w.completedAt`** on retry — prevents `isItemCompleted` false positives (new improvement over old code).
4. **`w.failedAt = ts()`** on final failure — proper timestamp (also new).
5. **Old code block removed** with a clear comment explaining the move.

### Tests

Excellent coverage — 5 source-pattern tests + 5 behavioral tests covering: ordering guarantee, retry advancement, maxRetries exhaustion, `source_plan` fallback scan, and happy path (PRD exists → done).

### Minor Notes (non-blocking)

- Guard changed from `meta?.item?.planFile` to `meta?.item?.id` — slightly broader but safe since inner logic degrades gracefully when `planFile` is absent.
- The try-catch around `mutateJsonFileLocked` is a good addition for resilience.

No issues found. Clean, well-tested fix.

---

yemi33:
## Build Fix

The CI failure on Node 18 is a **flaky pre-existing test** (`uid generates unique IDs: uid produced duplicates`) — a probabilistic uniqueness check on `shared.uid()` that intermittently fails due to timing on CI. Completely unrelated to this PR's changes.

**Evidence:**
- Node 20: ✅ pass
- Node 22: ✅ pass  
- Node 18: ❌ `uid generates unique IDs` (flaky — `crypto.randomBytes` timing collision)
- Local: ✅ 1474 passed, 0 failed

Re-triggered the failed Node 18 job.

Issue #779, opened 2026-04-10
engine: PRD item stays 'dispatched' (wip) after work item is deleted — no single source of truth


yemi33:
## Bug

When a work item is deleted directly from `work-items.json` (e.g. during cleanup of failed items), the corresponding `missing_features[].status` in the PRD JSON is not reset. The PRD retains `status: "dispatched"`, so the dashboard shows the item as **wip** indefinitely — even with no active agent, no dispatch entry, and no work item.

## Root cause

`syncPrdItemStatus` in `lifecycle.js` is the only mechanism that keeps PRD and work item status in sync. It is called on lifecycle events (dispatch, completion, failure) but **not on deletion**. Direct deletion of work items bypasses it entirely.

Additionally, `getPrdInfo` in `queries.js` uses:
```js
const rawStatus = wi ? (wi.status || item.status) : item.status;
```
When the work item is gone, it falls back to the (stale) PRD JSON value. There is no orphan detection.

## Design issue

The comment in `queries.js` says "PRD JSON status is the source of truth" but the code prefers work item status when available — so in practice **both are sources of truth** that can diverge. The system lacks a single authoritative source.

## Repro

1. Work item `P-xxx` is dispatched → PRD item synced to `dispatched`
2. Work item fails and is deleted from `work-items.json` directly
3. No lifecycle event fires → PRD item stays `dispatched`
4. Dashboard shows item as **wip** with no agent, no dispatch, no work item

## Fix applied locally (will be overwritten on next npm update)

`engine/queries.js` — auto-heal orphaned dispatches in `getPrdInfo`:
```js
// If PRD says dispatched but no work item exists, treat as pending (orphaned dispatch)
const rawStatus = wi ? (wi.status || item.status)
  : (item.status === 'dispatched' ? 'pending' : item.status);
```

## Proper fix needed in published package

Either:
- **Option A (minimal):** Apply the one-liner above to `getPrdInfo` + reset PRD status when deleting a work item via any API/mutation path
- **Option B (principled):** Make PRD JSON the single source of truth; derive display status from it alone; work item status is execution state only, not display state

## Affected items

`P-6b1e8d4f` (Notifications flyout) and `P-a6e1b5c4` (Backend: Task hub REST endpoint) — both showed as wip after cleanup.

yemi33:
[...]

PR #935, opened 2026-04-12
fix(queries): PRD updated status not overridden by work item done

### LINKED ISSUE #930: queries: work item "done" overrides PRD "updated" status, hiding rework signal in dashboard
## Bug

PRD items explicitly marked \`"updated"\` (plan scope changed, prior work needs rework) are displayed as \`"done"\` in the dashboard because \`queries.js\` unconditionally prefers work item status over PRD item status.

## Root Cause

In `engine/queries.js`, the status resolution loop treats work item status as the source of truth in all cases:

```js
// Before fix
const rawStatus = wi ? (wi.status || item.status) : ...;
```

When a PRD item is \`"updated"\` but a matching done work item exists (from the previous implementation), \`rawStatus\` becomes \`"done"\` — silently erasing the deliberate rework signal.

## Impact

- Dashboard shows 4 task hub items as \`DONE\` when they should show \`REDO\`
- User cannot see which items need rework vs which are genuinely complete
- The \`"updated"\` status in the PRD has no observable effect on the UI

## Fix

Treat \`"updated"\` as an intentional PRD-level override — do not let a done work item supersede it:

```js
// After fix (engine/queries.js)
const prdFlaggedForRework = item.status === PRD_ITEM_STATUS.UPDATED;
const rawStatus = (wi && !prdFlaggedForRework) ? (wi.status || item.status) : ...;
```

Also required adding \`PRD_ITEM_STATUS\` to the destructured imports in \`queries.js\`.

## Files

- \`engine/queries.js\` — status resolution loop (~line 960), imports (~line 14) — fixed
- Dashboard restart required for the module cache to pick up the change

## Related

- #928 — plan-to-prd playbook gap that causes \`"updated"\` items to be reset to \`"missing"\`
- #929 — syncPrdItemStatus has no backward-scan on PRD regeneration

yemi33:
## Summary

- **Fixes #930**: PRD items marked `updated` or `missing` (rework signals from plan resume / diff-aware PRD regeneration) were hidden in the dashboard because `getPrdInfo()` unconditionally preferred work item `done` status over the PRD item status.
- Added `prdFlaggedForRework` guard: when the PRD item is `updated` or `missing` AND the work item is in a done state (`DONE_STATUSES`), the PRD status now takes precedence in the display.
- Imported `PRD_ITEM_STATUS` and `DONE_STATUSES` constants into `queries.js`.

## Changes

- `engine/queries.js` — status resolution loop (~line 960): added rework-flag check before `rawStatus` computation; updated imports
- `test/unit.test.js` — source-pattern test verifying the fix uses correct constants

## Test plan

- [x] `npm test` — 1485 passed, 0 failed
- [ ] Manual: create a plan, complete items, edit plan to trigger diff-aware PRD update → verify dashboard shows `updated` not `done`

### DIFF
--- a/engine/queries.js
+++ b/engine/queries.js
@@ -11,7 +11,7 @@ const shared = require('./shared');
 
 const { safeRead, safeReadDir, safeJson, safeWrite, getProjects,
   projectWorkItemsPath, projectPrPath, parseSkillFrontmatter, KB_CATEGORIES,
-  WI_STATUS, ENGINE_DEFAULTS } = shared;
+  WI_STATUS, DONE_STATUSES, PRD_ITEM_STATUS, ENGINE_DEFAULTS } = shared;
 
 /**
  * Read the first `bytes` and last `bytes` of a file efficiently using byte offsets.
@@ -955,9 +955,12 @@ function getPrdInfo(config) {
   const statusDisplay = { pending: 'missing' };
   for (const item of items) {
     const wi = wiById[item.id];
-    // Work item status is source of truth when available (PRD JSON may lag behind)
+    // PRD 'updated'/'missing' = intentional rework signal — takes priority over a done work item (#930).
+    // Otherwise work item status is source of truth when available (PRD JSON may lag behind).
     // If PRD says dispatched/failed but no work item exists, treat as pending (orphaned — #779)
-    const rawStatus = wi ? (wi.status || item.status)
+    const prdFlaggedForRework = item.status === PRD_ITEM_STATUS.UPDATED || item.status === PRD_ITEM_STATUS.MISSING;
+    const rawStatus = (wi && !(prdFlaggedForRework && DONE_STATUSES.has(wi.status)))
+      ? (wi.status || item.status)
       : ((item.status === WI_STATUS.DISPATCHED || item.status === WI_STATUS.FAILED) ? WI_STATUS.PENDING : item.status);
     item.status = statusDisplay[rawStatus] || rawStatus || 'missing';
     // Attach execution metadata for display (agent, PR link, fail reason)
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### Root Cause
In `engine/queries.js` `getPrdInfo()`, the status resolution loop at ~line 960 unconditionally used work item status when a matching WI existed:
```js
const rawStatus = wi ? (wi.status || item.status) : ...;
```
When a PRD item was `updated` (rework signal from diff-aware regeneration) but the old work item was still `done`, the display showed `done` — hiding the rework signal entirely.

### Fix
Added a `prdFlaggedForRework` guard that checks whether the PRD item status is `updated` or `missing` (both are materializable statuses that signal rework). When this flag is true AND the work item is in a done state (`DONE_STATUSES`), the PRD status takes precedence:

```js
const prdFlaggedForRework = item.status === PRD_ITEM_STATUS.UPDATED || item.status === PRD_ITEM_STATUS.MISSING;
const rawStatus = (wi && !(prdFlaggedForRework && DONE_STATUSES.has(wi.status)))
  ? (wi.status || item.status)
  : ...;
```

This preserves the existing behavior for all other cases:
- WI `dispatched`/`pending`/`failed` still override PRD status (active execution)
- No WI + orphaned PRD `dispatched`/`failed` → reset to `pending` (#779 fix preserved)
- PRD `done` + WI `done` → shows `done` (normal completion)

### Constants Added to Imports
- `PRD_ITEM_STATUS` — for checking `UPDATED` and `MISSING`
- `DONE_STATUSES` — for checking if WI is in a terminal done state (includes legacy aliases)

### Tests
- 1485 passed, 0 failed
- Added source-pattern test verifying the fix uses `PRD_ITEM_STATUS.UPDATED`, `PRD_ITEM_STATUS.MISSING`, `DONE_STATUSES.has`, and `prdFlaggedForRework`

PR #940, opened 2026-04-12
fix(lifecycle): syncPrdItemStatus backward-scan for PRD reset

### LINKED ISSUE #929: engine: syncPrdItemStatus has no backward-scan on PRD regeneration — done work items cannot correct reset statuses
## Bug

When a PRD file is regenerated wholesale (plan-to-prd), any status corrections that \`syncPrdItemStatus\` would have written are lost. If a regeneration incorrectly sets an item to \`"missing"\` when a \`done\` work item exists for that ID, the engine has no mechanism to detect and fix this.

## Root Cause

\`syncPrdItemStatus\` in \`engine/lifecycle.js\` fires **reactively** — only when a work item changes state (completes, PR merges). It does not scan existing work items against PRD items proactively.

If a PRD is rewritten with wrong statuses and no work item state-change occurs afterward, the correction never fires.

## Scenario That Exposed This

1. v2 PRD (officeagent-2026-04-09 era) — items correctly marked \`"updated"\`
2. v2 PRD status \`awaiting-approval\` — no work items materialized, so no \`syncPrdItemStatus\` triggers existed for those IDs
3. v3 PRD regenerated — Lambert reset those items to \`"missing"\`
4. Engine ticks — no existing done work items for those IDs → no corrective sync fired
5. Result: items stuck as \`"missing"\` even though prior branches existed

## Proposed Fix

In the engine's per-tick cleanup (or on PRD file creation detection), add a scan:

```js
// For each PRD item marked "missing", check if a done work item exists for that ID.
// If found, promote PRD item to "updated" (prior work exists, needs revision).
for (const feature of plan.missing_features) {
  if (feature.status === 'missing') {
    const wi = allWorkItems.find(w => w.id === feature.id && DONE_STATUSES.has(w.status));
    if (wi) {
      feature.status = 'updated';
      // write back
    }
  }
}
```

This provides a safety net that catches both agent errors and race conditions where a work item completes while the PRD is being rewritten.

## Files

- \`engine/lifecycle.js\` — \`syncPrdItemStatus\` (reactive only, no proactive scan)
- \`engine.js\` — tick cycle, could add a reconcile step

yemi33:
## Summary

- Adds `reconcilePrdStatuses()` in `engine/lifecycle.js` — a proactive backward-scan that promotes "missing" PRD items to "updated" when a done work item already exists for that ID
- Called before `materializePlansAsWorkItems()` in each discovery tick, ensuring PRD statuses are corrected before materialization processes them
- Skips completed PRDs; uses O(1) lookup via Map for done work items; per-file try-catch for fault isolation

Closes #929

## Why

When a PRD is regenerated, the plan-to-prd agent may incorrectly mark items as "missing" when done work items already exist. The existing `syncPrdItemStatus` is purely reactive (fires only on WI state changes), so these mismatches persist silently. The materializer treats "missing" differently from "updated" — "updated" dispatches to the existing branch while "missing" may create from scratch. This backward-scan ensures the correct semantics.

## Test plan

- [x] `reconcilePrdStatuses` exported from lifecycle.js
- [x] Called before `materializePlansAsWorkItems` in `discoverWork` 
- [x] Source-pattern tests verify correct constant usage (PRD_ITEM_STATUS, DONE_STATUSES, PLAN_STATUS)
- [x] No-crash test for empty/missing PRD directory
- [x] All 1488 unit tests pass (0 failures)

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -135,7 +135,7 @@ const { renderPlaybook, validatePlaybookVars, PLAYBOOK_REQUIRED_VARS,
 
 // ─── Lifecycle (extracted to engine/lifecycle.js) ────────────────────────────
 
-const { runPostCompletionHooks, updateWorkItemStatus, syncPrdItemStatus, handlePostMerge, checkPlanCompletion,
+const { runPostCompletionHooks, updateWorkItemStatus, syncPrdItemStatus, reconcilePrdStatuses, handlePostMerge, checkPlanCompletion,
   syncPrsFromOutput, updatePrAfterReview, updatePrAfterFix, checkForLearnings, extractSkillsFromOutput,
   updateAgentHistory, updateMetrics, createReviewFeedbackForAuthor, parseAgentOutput, syncPrdFromPrs,
   isItemCompleted, classifyFailure, processPendingRebases } = require('./engine/lifecycle');
@@ -2692,6 +2692,7 @@ async function discoverWork(config) {
 
   // Side-effect passes: materialize plans and design docs into work-items.json
   // These write to project work queues — picked up by discoverFromWorkItems below.
+  reconcilePrdStatuses(config); // Backward-scan: correct "missing" PRD items that have done work items (#929)
   materializePlansAsWorkItems(config);
 
   for (const project of projects) {
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -8,7 +8,7 @@ const path = require('path');
 const os = require('os');
 const shared = require('./shared');
 const { safeRead, safeJson, safeWrite, mutateJsonFileLocked, mutateWorkItems, execSilent, execAsync, projectPrPath, getPrLinks, addPrLink,
-  log, ts, dateStamp, WI_STATUS, DONE_STATUSES, PLAN_TERMINAL_STATUSES, WORK_TYPE, PLAN_STATUS, PR_STATUS, DISPATCH_RESULT,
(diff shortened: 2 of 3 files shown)


yemi33:
## Implementation Notes

### Architecture Decision

Added `reconcilePrdStatuses()` as a **separate function** rather than modifying `syncPrdItemStatus` directly. Rationale:

- `syncPrdItemStatus` is reactive (item-scoped, event-driven) — changing it to also scan all PRDs would violate SRP
- The backward-scan is a reconciliation concern that runs on a schedule (per-tick), not on-demand
- Separating allows independent throttling if needed (currently runs every tick; could easily be gated to every N ticks)

### Data Flow

```
discoverWork()
  → reconcilePrdStatuses(config)    ← NEW: promotes missing→updated for done WIs
  → materializePlansAsWorkItems()   ← existing: re-opens WIs based on PRD status
```

The backward-scan runs **before** materialization so the materializer always sees corrected statuses. This means "updated" items get dispatched to their existing branches rather than being treated as new work.

### Performance

- O(n) scan of all work items to build a Map of done WIs (once per tick)
- O(m) scan of PRD files (typically 0-10 files)
- Early exits: no PRD dir, no work items, no done work items, completed PRDs
- Per-file try-catch: one bad PRD file doesn't block reconciliation of others

### Files Changed

| File | Change |
|------|--------|
| `engine/lifecycle.js` | Added `reconcilePrdStatuses()`, imported `PRD_ITEM_STATUS` |
| `engine.js` | Imported and called `reconcilePrdStatuses(config)` before materialization |
| `test/unit.test.js` | 5 new tests (export check, call ordering, source pattern, no-crash) |

PR #966, opened 2026-04-13
feat(P-c7d3e1a6): fix dispatch.js mutator fallback to use nullish coalescing

yemi33:
## Summary
- Changes `mutator(dispatch) || dispatch` to `mutator(dispatch) ?? dispatch` in `mutateDispatch()` (engine/dispatch.js:33)
- The `||` operator incorrectly treats any falsy mutator return (0, false, '') as "no return" and falls back to `dispatch`. The `??` operator only falls back when the mutator returns `null` or `undefined`, which is the correct behavior for in-place mutation patterns.
- Adds 3 unit tests: source pattern check for `??`, behavioral test for undefined return fallback, and behavioral test for explicit return value usage

## Test plan
- [x] Source pattern test verifies `??` is used, not `||`
- [x] Behavioral test: mutator returning `undefined` (in-place mutation) falls back to dispatch
- [x] Behavioral test: mutator returning a dispatch object uses the returned value
- [x] All 1241 existing tests pass with 0 failures

### DIFF
--- a/engine/dispatch.js
+++ b/engine/dispatch.js
@@ -30,7 +30,7 @@ function mutateDispatch(mutator) {
     dispatch.pending = Array.isArray(dispatch.pending) ? dispatch.pending : [];
     dispatch.active = Array.isArray(dispatch.active) ? dispatch.active : [];
     dispatch.completed = Array.isArray(dispatch.completed) ? dispatch.completed : [];
-    return mutator(dispatch) || dispatch;
+    return mutator(dispatch) ?? dispatch;
   }, { defaultValue: defaultDispatch });
   // Invalidate the read cache so next getDispatch() sees fresh data
   try { require('./queries').invalidateDispatchCache(); } catch {}
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -2424,6 +2424,56 @@ async function testStateIntegrity() {
       'engine dispatch writes should use lock-backed mutation');
   });
 
+  await test('mutateDispatch uses nullish coalescing (??) not OR (||) for mutator fallback', () => {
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

### Change
Single-character fix in `engine/dispatch.js:33`: `||` → `??`

**Before:** `return mutator(dispatch) || dispatch;`
**After:** `return mutator(dispatch) ?? dispatch;`

### Why
The `||` (OR) operator falls back to `dispatch` for **any** falsy return value: `0`, `false`, `''`, `null`, `undefined`. The intent is to fall back only when the mutator doesn't return anything (in-place mutation pattern — returns `undefined`). The `??` (nullish coalescing) operator only falls back for `null` or `undefined`, correctly preserving any other returned value.

### Impact
In practice, all existing `mutateDispatch` callbacks return the dispatch object, so there's no behavioral change for current code. But this makes the contract correct: future mutators could theoretically return other values without being silently overridden.

### Tests Added
1. **Source pattern test**: Verifies `??` is used, `||` is not
2. **Behavioral: undefined return**: Mutator with no return (in-place) → dispatch object persisted correctly
3. **Behavioral: explicit return**: Mutator returning dispatch object → value used directly

All 1241 existing tests continue to pass.

yemi33 (commented):
VERDICT: APPROVE

Clean, correct one-liner fix: `||` → `??` in `mutateDispatch` fallback (dispatch.js:33). The `||` operator would incorrectly discard falsy-but-valid returns; `??` only falls back on `null`/`undefined`, which is the right semantic for in-place mutators that don't explicitly return.

**Tests:** Three new tests cover: (1) source-code assertion for `??`, (2) behavioral test for undefined return (in-place mutation), (3) behavioral test for explicit return. All well-structured.

No issues found.

PR #980, opened 2026-04-13
fix(W-mnwnfsxkp143): prevent Create Plan from meeting saving doc-chat context bleed

### LINKED ISSUE #961: "Create Plan" from meeting saves doc-chat explanation instead of plan content
## Problem

The "Create Plan" button on the meeting detail view calls \`/api/doc-chat\` to generate plan content, then blindly saves whatever the response is to \`plans/\` via \`/api/plans/create\`.

If the doc-chat session already has context from a prior CC conversation, it returns a *summary of what was already done* rather than a new plan. The dashboard saves this explanation text as a plan file — resulting in garbage like:

> "Here's the plan, written to \`plans/engine-correctness-and-perf-2026-04-12.md\` and ready to submit for approval..."

## Root Cause

\`/api/plans/create\` accepts whatever doc-chat returns without validating it looks like actual plan content. The doc-chat session bleeds context from prior CC conversations, causing it to summarise past work instead of drafting a new plan.

## Fix

Two options (either or both):

1. **Validate content on save** — in \`/api/plans/create\`, reject content that doesn't look like a plan (e.g. doesn't start with a `#` heading, or references an existing plan file path). Return a 400 with a clear error so the UI can surface it.

2. **Isolate the doc-chat call** — in \`render-meetings.js\` ~line 435, use a fresh session (no \`sessionId\`) for the plan-generation doc-chat call so it can't bleed context from unrelated CC conversations.

Option 2 is the cleaner fix — the root cause is session contamination, not bad content.

## Repro

1. Have an active CC session where a plan was discussed
2. Open a meeting detail modal
3. Click "Create Plan"
4. Doc-chat returns a summary of the prior CC conversation; the file saved to \`plans/\` contains that summary instead of a plan

yemi33:
## Summary

- **Root cause fix**: Add `freshSession` support to `ccDocCall` — when true, deletes existing session and skips resume, ensuring the plan-generation doc-chat call starts with a clean context and cannot bleed from prior conversations
- **Frontend guard**: Validate generated content looks like a plan (starts with `#`, `**`, or list marker) before saving — rejects obvious meta-responses
- **Backend guard**: `/api/plans/create` rejects content that doesn't start with a markdown heading, bold text, or list item

Closes #961

## Test plan
- [x] 6 new unit tests added and passing (source-code string matching)
- [x] Full test suite passes (1243 PASS, 0 failures)
- [ ] Manual: open meeting detail → chat with Q&A → click Create Plan → verify it generates fresh plan content, not a CC summary
- [ ] Manual: click Create Plan twice on same meeting → second call also generates fresh content (no session reuse)

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -792,7 +792,7 @@ async function ccDocCall({ message, document, title, filePath, selection, canEdi
   // bleed context from earlier conversations.
   if (freshSession && sessionKey) {
     docSessions.delete(sessionKey);
-    persistDocSessions();
+    // Skip persistDocSessions() here — the post-call cleanup below handles persistence.
   }
 
   // Skip re-sending full document on session resume if content unchanged
--- a/dashboard.js
+++ b/dashboard.js
@@ -783,13 +783,21 @@ async function ccCall(message, { store = 'cc', sessionKey, extraContext, label =
 }
 
 // Doc-specific wrapper — adds document context, parses ---DOCUMENT---
-async function ccDocCall({ message, document, title, filePath, selection, canEdit, isJson, model }) {
+async function ccDocCall({ message, document, title, filePath, selection, canEdit, isJson, model, freshSession }) {
   const sessionKey = filePath || title;
   const docSlice = document.slice(0, 20000);
 
+  // freshSession: true → discard any prior session for this key so the call starts clean.
+  // Used by one-shot generation flows (e.g. Create Plan from meeting) that must not
+  // bleed context from earlier conversations.
+  if (freshSession && sessionKey) {
+    docSessions.delete(sessionKey);
+    persistDocSessions();
+  }
+
   // Skip re-sending full document on session resume if content unchanged
   const docHash = require('crypto').createHash('md5').update(docSlice).digest('hex').slice(0, 8);
-  const existing = resolveSession('doc', sessionKey);
+  const existing = freshSession ? null : resolveSession('doc', sessionKey);
   const docUnchanged = existing?.sessionId && existing._docHash === docHash;
 
   let docContext;
@@ -808,8 +816,14 @@ async function ccDocCall({ message, document, title, filePath, selection, canEdi
     skipStatePreamble: true,
     ...(model ? { model } : {}),
   });
-  // Store doc hash for next call's unchanged check
-  if (result.code === 0 && result.sessionId) {
+
+  if (freshSession && sessionKey) {
+    // One-shot call — discard the session ccCall just stored so it cannot
+    // bleed into future interactions under the same key.
+    docSessions.delete(sessionKey);
+    persistDocSessions();
+  } else if (result.code === 0 && result.sessionId) {
+    // Store doc hash for next call's unchanged check
     const session = resolveSession('doc', sessionKey);
     if (session) session._docHash = docHash;
   }
@@ -2819,6 +2833,7 @@ What would you like to discuss or change? When you're happy, say "approve" and I
         message: body.message, document: currentContent, title: body.title,
         filePath: body.filePath, selection: body.selection, canEdit, isJson,
         model: body.model || undefined,
+        freshSession: !!body.freshSession,
       });
 
       if (!content) return jsonReply(res, 200, { ok: true, answer, edited: false, actions });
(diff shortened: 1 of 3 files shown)


yemi33:
## Implementation Notes

### Root Cause Analysis

The `_createPlanFromMeeting` function in `render-meetings.js` calls `/api/doc-chat` with `title: 'Meeting: ' + m.title` but no `filePath`. The backend session key for this call becomes the title string. If a prior doc-chat call (from a previous Create Plan click, or from the Q&A panel using the same title) created a session under that key, `ccDocCall` would resume that session. The LLM, with context from the prior conversation, would respond with a summary of what it already did rather than generating fresh plan content.

### Changes Made

**`dashboard/js/render-meetings.js`**
- Added `freshSession: true` to the doc-chat request body at the plan-generation call (line ~437)
- Added content validation: rejects generated text that doesn't start with `#`, `**`, or `- ` (list marker), catching meta-responses before they reach `/api/plans/create`

**`dashboard.js`**
- `handleDocChat`: Threads `body.freshSession` to `ccDocCall`
- `ccDocCall`: New `freshSession` parameter — when true:
  1. Deletes any existing session for the key from `docSessions`
  2. Passes `null` instead of the resolved session to `ccCall` (forces fresh session)
  3. After `ccCall` returns, deletes the newly-stored session so it doesn't pollute future calls
- `/api/plans/create`: Added content validation — rejects content that doesn't start with a markdown heading, bold text, or list item

**`test/unit.test.js`**
- 6 new tests validating `freshSession` plumbing and content validation

### Design Decisions

1. **Session cleanup after one-shot call**: The `freshSession` path deletes the session both before and after `ccCall`. The "before" prevents resume; the "after" prevents the newly-created session from polluting future one-shot calls on the same meeting.

2. **Two-layer content validation**: Frontend checks the generated content before sending to `/api/plans/create`, and the backend also validates. Belt-and-suspenders — the frontend gives a better UX, the backend prevents bad data regardless of caller.

3. **Regex `^[#*\-]` for plan detection**: Intentionally permissive — accepts headings (`#`), bold text (`**`), and list items (`- ` or `* `). This allows diverse plan formats while catching obvious non-plan meta-responses that typically start with "I" or "The" or similar prose.

yemi33 (commented):
VERDICT: APPROVE

## Review: PR-980 — prevent Create Plan from meeting saving doc-chat context bleed

### Summary

Clean, well-scoped fix for #961. The `freshSession` parameter in `ccDocCall` correctly prevents session reuse for one-shot generation flows. The two-layer content validation (frontend + backend) adds defense-in-depth.

### What's good

- **Root cause fix is correct**: Deleting the session before *and* after `ccCall` handles both directions of bleed — no stale context on entry, no leaked session on exit.
- **`freshSession ? null : resolveSession` is clean**: Skips both session resume and the `docUnchanged` optimization, ensuring the full document is always sent.
- **`!!body.freshSession` coercion**: Properly sanitizes the input to boolean.
- **Tests cover the plumbing end-to-end**: All 6 tests verify the `freshSession` flag is threaded from frontend → handler → `ccDocCall`.

1. **Frontend/backend regex inconsistency**: The frontend regex `/^[#*\-]/` accepts `*foo` (single asterisk), while the backend requires `**` (bold) or `* ` (list with space). A plan starting with `*italic*` would pass the frontend but fail the backend with a 400 error. Not a real concern — the LLM prompt says "start with the plan title" which implies a heading — but the error message would be confusing to a user if it ever hit. Consider aligning the frontend check to match the backend: `/^(#|\*\*|[-*] )/.test(...)`.

2. **Double `persistDocSessions()` on fresh calls**: Called both before and after `ccCall`. Each writes to disk. The "before" write could be deferred — just `docSessions.delete()` in memory and let the "after" write handle persistence. Negligible impact since plan generation is infrequent.

3. **Source-code inspection tests**: All 6 tests use string matching against source code. This is the established pattern in this codebase for plumbing validation, and behavioral tests would require mocking the LLM pipeline, so this is the right tradeoff here.

### Verdict

The fix correctly addresses the root cause, adds appropriate guardrails, and has good test coverage. Ship it.

yemi33:
## Review Feedback Addressed

Addressed the two non-blocking observations from Ripley's review:

1. **Frontend/backend regex aligned**: Changed frontend check from `/^[#*\-]/` to `/^(#|\*\*|[-*] )/` — now both layers reject the same edge cases (e.g. `*italic*` without space).

2. **Removed redundant `persistDocSessions()`**: The pre-call delete now only clears the in-memory Map. The post-call cleanup (which always runs) handles disk persistence.

**Tests:** 1243 passed, 0 failures (no changes needed — existing tests still pass with aligned regex).

PR #1000, opened 2026-04-13
fix(W-mnxomrllro4i): error_max_turns misclassified as permission-blocked

### LINKED ISSUE #995: error_max_turns misclassified as permission-blocked when startup hook also fails
## Summary

When an agent hits the turn limit (`error_max_turns`) AND has a startup hook that exited with a non-zero code, the engine classifies the run as `permission-blocked` instead of `max-turns`. The failure class should reflect the primary stop reason (`error_max_turns`), not the hook exit code.

## Observed

Three fix agents all show `failureClass: "permission-blocked"` in dispatch.json, but their output logs all end with:
```
{"type":"result","subtype":"error_max_turns",...,"is_error":true,"num_turns":51}
```

All three sessions also have a `SessionStart` hook exiting with code 28 (curl operation timeout), which appears to be what the classification picks up.

## Root cause

`classifyFailure()` in `lifecycle.js` (around line 1886) correctly detects `error_max_turns`:
```js
if (/error_max_turns|terminal_reason.*max_turns|max.*turns.*reached/i.test(combined)) {
  return FAILURE_CLASS.MAX_TURNS;
}
```

However, the hook failure (exit_code 28 on SessionStart) is likely matched by an earlier branch in the classification logic � specifically the `permission-blocked` pattern � which short-circuits before the `error_max_turns` check is reached. Since the agent ran all 51 turns after the hook failure, the hook exit is not the real failure cause.

## Fix

The `error_max_turns` check should run **before** any hook-exit-based classification. If the result subtype is `error_max_turns`, that is definitively the stop reason regardless of hook exit codes. Hook failures at startup are informational context, not the terminal cause, when the agent subsequently ran to turn exhaustion.

## Affected runs
- `ralph-fix-mnxk06xq748m` � 51 turns, PR-5055109 merge conflict
- `ralph-fix-mnxle13pln6b` � 51 turns, PR-5055109 merge conflict (retry)
- `rebecca-fix-mnxle14gnp9s` � 51 turns, PR-5055985 merge conflict

## Impact

Misclassification affects retry/recovery logic � `max-turns` and `permission-blocked` may trigger different retry strategies. Also misleads debugging (the actual problems were ADO auth token exhaustion and a pre-existing `bebop-ux` typecheck failure, not permission issues).

---
_Filed via Minions dashboard_

yemi33:
## Summary

- **Reorders `classifyFailure()`** so `error_max_turns` is checked immediately after `CONFIG_ERROR` (exit code 78), before `permission-blocked` and all other pattern matches
- Fixes misclassification where hook failures (SessionStart curl timeout exit code 28) injected permission/auth patterns into stderr, causing `classifyFailure()` to return `PERMISSION_BLOCKED` instead of `MAX_TURNS` — even when the agent ran all 51 turns
- This prevented retries (`PERMISSION_BLOCKED` has `NO_RETRY` policy, while `MAX_TURNS` has `RETRY_SAME`)

## Root Cause

In `classifyFailure()` (lifecycle.js ~line 1886), the permission-blocked regex (`/permission denied|access denied|unauthorized|.../`) fired before the `error_max_turns` check. Hook startup failures write auth-related error text to stderr, but the agent subsequently ran to turn exhaustion — `error_max_turns` is the definitive stop reason from Claude CLI.

Affected runs: ralph-fix-mnxk06xq748m (51 turns), ralph-fix-mnxle13pln6b (51 turns), rebecca-fix-mnxle14gnp9s (51 turns).

## Test plan

- [x] Added regression test: `error_max_turns` wins over `permission-blocked` when both patterns present in combined output
- [x] Tests 3 variants: `permission denied`, `access denied`, `unauthorized` — all correctly yield `MAX_TURNS`
- [x] All 1536 unit tests pass, 0 failures

Fixes #995

### DIFF
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -1883,6 +1883,15 @@ function classifyFailure(code, stdout = '', stderr = '') {
   // Exit code 78 — configuration error (Claude CLI not found, bad setup)
   if (code === 78) return FAILURE_CLASS.CONFIG_ERROR;
 
+  // Max turns exhausted (error_max_turns) — work in progress, retryable
+  // MUST be checked FIRST (before permission-blocked, out-of-context, etc.)
+  // because hook failures (e.g. SessionStart curl timeout exit code 28) can inject
+  // permission/auth patterns into stderr even when the agent ran to turn exhaustion.
+  // The error_max_turns result subtype is the definitive stop reason from Claude CLI.
+  if (/error_max_turns|"subtype"\s*:\s*"error_max_turns"|terminal_reason.*max_turns|max.*turns.*reached/i.test(combined)) {
+    return FAILURE_CLASS.MAX_TURNS;
+  }
+
   // Permission / trust / auth failures
   if (/permission denied|access denied|unauthorized|403 forbidden|trust.*blocked|auth.*fail/i.test(combined)) {
     return FAILURE_CLASS.PERMISSION_BLOCKED;
@@ -1891,12 +1900,6 @@ function classifyFailure(code, stdout = '', stderr = '') {
   // Merge conflicts
   if (/merge conflict|conflict.*merge|automatic merge failed|fix conflicts/i.test(combined)) {
     return FAILURE_CLASS.MERGE_CONFLICT;
-  }
-
-  // Max turns exhausted (error_max_turns) — work in progress, retryable
-  // Must be checked BEFORE OUT_OF_CONTEXT to avoid misclassification as non-retryable
-  if (/error_max_turns|"subtype"\s*:\s*"error_max_turns"|terminal_reason.*max_turns|max.*turns.*reached/i.test(combined)) {
-    return FAILURE_CLASS.MAX_TURNS;
   }
 
   // Context window exhausted (token limit, context length — NOT max turns)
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### What changed
1. **`engine/lifecycle.js`** — Moved the `error_max_turns` check block from after `permission-blocked` to immediately after `CONFIG_ERROR` (exit code 78). The `error_max_turns` result subtype from Claude CLI is a definitive stop reason and must take priority over any pattern-matched heuristics.

2. **`test/unit.test.js`** — Added regression test `classifyFailure: error_max_turns wins over permission-blocked hook noise (#995)` with three assertions covering the exact scenario: `error_max_turns` JSON in stdout combined with `permission denied`, `access denied`, and `unauthorized` patterns in stderr from hook failures.

### Why this ordering matters
- `PERMISSION_BLOCKED` → `NO_RETRY` escalation policy (never retried)
- `MAX_TURNS` → `RETRY_SAME` escalation policy (retryable on same agent)

Misclassification as `permission-blocked` meant agents that ran 51 turns of productive work were treated as blocked instead of retried.

### Design principle
The `error_max_turns` / `max_turns_reached` result subtype is emitted by Claude CLI as a structured result event. It's not a heuristic pattern match — it's a definitive signal. Heuristic pattern matches (permission, auth, trust) should never override it.

yemi33:
Duplicate of #1001 (merged). Same fix: move MAX_TURNS check before PERMISSION_BLOCKED in classifyFailure(). Closing in favor of #1001 which has better-organized regression tests.

PR #1015, opened 2026-04-13
fix(W-mnxu9t8uascs): steering no-session kill re-queues dispatch

### LINKED ISSUE #1014: steering kill on no-session doesn't re-queue dispatch item — task errors instead of retrying
## Bug

When a steer message arrives before an agent has established a `sessionId` (very early in startup), `timeout.js` kills the agent and forwards the message to the agent's inbox with the log message *"your message will be delivered on retry."* However, the dispatch item is **not re-queued** — it ends up in `error` status and is never retried. The inbox message sits undelivered indefinitely.

## Reproduction

1. Dispatch an agent to a task
2. Within the first few seconds (before the Claude session is established), send a steer message
3. Observe: agent is killed, status → `error`, steer message lands in inbox but no retry occurs

## Root Cause

`timeout.js:92–111` handles the no-session case: kills the process and saves the message to inbox, but does **not** re-queue the dispatch item. The close event handler then marks it as `error`. The "delivered on retry" message is misleading — it only holds if the user manually retries.

## Expected Behavior

One of:
- **Option A:** When `_steeringNoSession` causes a kill, the close handler should re-queue the dispatch item (not mark as `error`) so the engine retries it and picks up the inbox message
- **Option B:** Delay steering delivery — if the agent is very young (< ~30s), wait for session establishment before acting on the steer instead of killing immediately

## Observed Log

```
[steering] Message received but no session to resume. Killing agent — your message will be delivered on retry.
```

Agent status: `error`  
Dispatch: not re-queued  
Inbox: `steering-<ts>.md` present but never consumed

yemi33:
## Summary

- When a steer message arrives before an agent has established a sessionId, `timeout.js` kills the agent and saves the message to inbox — but previously the close handler treated this as a normal error, consuming a retry count and eventually marking the work item as failed
- The close handler in `engine.js` now detects the `_steeringNoSession` flag and moves the dispatch item back to pending instead of completing it as error, so the engine retries and the agent picks up the inbox message on the next dispatch

Closes #1014

## Test plan

- [x] 3 new unit tests verify the close handler detects `_steeringNoSession`, re-queues to pending (not error), cleans up process maps, and logs the action
- [x] All 1670 tests pass, 0 failures

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -1098,6 +1098,34 @@ async function spawnAgent(dispatchItem, config) {
       return;
     }
 
+    // Check if this was a no-session steering kill (#1014) — re-queue for retry instead of erroring.
+    // timeout.js sets _steeringNoSession when it kills an agent that hasn't established a sessionId.
+    // The steering message is already saved to inbox, so re-queuing lets the engine retry and the
+    // agent picks up the message on the next dispatch.
+    if (procInfo?._steeringNoSession) {
+      log('info', `Steering no-session: re-queue ${agentId} (${id}) — dispatch moved back to pending`);
+      activeProcesses.delete(id);
+      realActivityMap.delete(id);
+      // Move dispatch item from active back to pending so engine retries
+      mutateDispatch((dp) => {
+        const idx = dp.active.findIndex(d => d.id === id);
+        if (idx >= 0) {
+          const item = dp.active.splice(idx, 1)[0];
+          delete item.started_at;
+          delete item.workerState;
+          delete item.workerStateAt;
+          delete item.workerStateDetail;
+          dp.pending.push(item);
+        }
+        return dp;
+      });
+      // Cleanup temp files
+      try { fs.unlinkSync(sysPromptPath); } catch { /* cleanup */ }
+      try { fs.unlinkSync(promptPath); } catch { /* cleanup */ }
+      try { fs.unlinkSync(promptPath.replace(/prompt-/, 'pid-').replace(/\.md$/, '.pid')); } catch { /* cleanup */ }
+      return;
+    }
+
     activeProcesses.delete(id);
     realActivityMap.delete(id); // Clean up real activity tracking
 
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -6955,6 +6955,43 @@ async function testAgentSteering() {
       'Should write confirmation to live-output.log so user sees it in dashboard');
   });
 
+  // ── Steering no-session re-queue (#1014) ─────────────────────────────────────
+
+  await test('close handler detects _steeringNoSession and re-queues dispatch instead of erroring', () => {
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

### Root Cause
`timeout.js:92–111` handles the no-session steering case by killing the agent and saving the steering message to inbox. It sets `_steeringNoSession = true` on the process info — but the close handler in `engine.js` (`onAgentClose`) didn't check this flag. It fell through to the normal error completion path, calling `completeDispatch(id, ERROR, ...)` which incremented the retry counter and eventually exhausted retries.

### Fix
Added a check for `procInfo?._steeringNoSession` in `onAgentClose` (engine.js:1100–1130), positioned after the `_steeringMessage` check but before the normal completion path. When detected:

1. Logs the re-queue action for observability
2. Removes the process from `activeProcesses` and `realActivityMap`
3. Uses `mutateDispatch()` to atomically move the dispatch item from `active` back to `pending` (clearing stale worker state fields)
4. Cleans up temp files (prompt, sysprompt, PID file)
5. Returns early — skips normal error completion entirely

This means the dispatch item returns to the pending queue without consuming a retry, and on the next engine tick, the agent is re-dispatched and picks up the steering message from its inbox.

### Files Changed
- `engine.js` — 28-line insertion in `onAgentClose` close handler
- `test/unit.test.js` — 3 new source-analysis tests (TDD: tests written first, verified failing, then implementation added)

yemi33 (commented):
VERDICT: APPROVE

Clean, well-placed fix. The `_steeringNoSession` check is correctly positioned in the close handler — after the steering-resume path but before the normal completion flow. Key observations:

1. **Correct dispatch mutation** — uses `mutateDispatch()` for atomic read-modify-write on dispatch.json, consistent with concurrency rules.
2. **Proper state cleanup** — removes from `activeProcesses` and `realActivityMap` before re-queuing, preventing stale entries.
3. **Transient field stripping** — correctly removes `started_at`, `workerState`, `workerStateAt`, `workerStateDetail` so the re-queued item looks fresh to the dispatcher.
4. **No retry counter increment** — appropriate since this isn't a failure; the agent never got a chance to run.
5. **Temp file cleanup** — consistent with other early-return paths in the same function.
6. **Tests** — source-code string matching, consistent with existing test patterns in the steering test section.

No issues found. The flow is: timeout.js sets `_steeringNoSession` → kills agent → close handler detects flag → re-queues to pending → engine retries on next tick with steering message in inbox.

yemi33:
## Fix Review Status

No review issues to address — PR-1015 was approved by Ripley (Lead/Explorer) with VERDICT: APPROVE and no findings requiring changes. The only comment on the PR is the original implementation notes from the author.

All 1670 tests pass, 0 failures. No code changes needed.

PR #36, opened 2026-04-01
feat(P-8e7f6a5b): eval iteration tracking + needs-human-review terminal status

yemi33:
## Summary

- **Eval iteration loop**: When an evaluate work item completes with `pass: false`, the engine parses the structured verdict, increments `_evalIterations` on the parent implement item, and auto-creates a `fix` work item with evaluator feedback (`_evalFeedback`, `_evalCriteriaFailed`) injected into the fix context
- **Terminal escalation**: When `_evalIterations >= evalMaxIterations` (default 3), the parent is set to `needs-human-review` instead of spawning another fix — engine will not dispatch these items
- **Dashboard**: New amber/orange `needs-review` badge, PRD item row border style, and edit/archive/feedback actions for `needs-human-review` items
- **`parseEvalVerdict()`**: Extracts structured JSON verdict from evaluate agent output (supports fenced and bare JSON)
- **8 new unit tests**: Verdict parsing, iteration counting, escalation at max, fix creation below max, terminal status gating, dashboard rendering

## Depends On

- PR-33 (auto-dispatch evaluate after implement)
- PR-29 (evaluate playbook + routing)

## Test plan

- [x] 591 tests pass, 0 failures (8 new tests added)
- [ ] Verify `needs-human-review` badge renders amber/orange in dashboard
- [ ] End-to-end: implement → evaluate (fail) → fix → evaluate (fail) → fix → evaluate (fail) → needs-human-review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1019, opened 2026-04-14
fix(W-mnxw6zvwaof6): add unhandledRejection/uncaughtException handlers to engine process

yemi33:
## Summary

- Adds `process.on('unhandledRejection')` and `process.on('uncaughtException')` handlers in `engine/cli.js` near the existing SIGTERM/SIGINT handlers
- Each handler logs the error via `shared.log('fatal', ...)`, flushes buffered logs via `shared.flushLogs()`, then exits with code 1
- All logging/flushing is wrapped in try-catch to ensure the process exits even if logging itself fails

## Problem

The engine process had no crash handlers. Any unhandled promise rejection or uncaught exception would kill the Node process silently — no log entry, no restart signal, the tick loop just stops.

## Test plan

- [x] 4 new source-inspection tests verifying handler registration, logging, flushing, and exit(1)
- [x] All 1682 tests pass (1 pre-existing unrelated failure)

### DIFF
--- a/engine/cli.js
+++ b/engine/cli.js
@@ -493,6 +493,23 @@ const commands = {
 
     process.on('SIGTERM', () => gracefulShutdown('SIGTERM'));
     process.on('SIGINT', () => gracefulShutdown('SIGINT'));
+
+    // Crash handlers — log the error and exit cleanly so the process is detectable as crashed
+    process.on('unhandledRejection', (reason) => {
+      const msg = reason instanceof Error ? reason.stack || reason.message : String(reason);
+      console.error(`[FATAL] Unhandled promise rejection: ${msg}`);
+      try { shared.log('fatal', `Unhandled promise rejection: ${msg}`); } catch { /* best effort */ }
+      try { shared.flushLogs(); } catch { /* best effort */ }
+      process.exit(1);
+    });
+
+    process.on('uncaughtException', (err) => {
+      const msg = err instanceof Error ? err.stack || err.message : String(err);
+      console.error(`[FATAL] Uncaught exception: ${msg}`);
+      try { shared.log('fatal', `Uncaught exception: ${msg}`); } catch { /* best effort */ }
+      try { shared.flushLogs(); } catch { /* best effort */ }
+      process.exit(1);
+    });
   },
 
   stop() {
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -10491,6 +10491,39 @@ async function testAuxModuleBugFixes() {
     assert.ok(src.includes('fs.unlinkSync(sysTmpPath)'), 'Cleanup should delete sysTmpPath');
   });
 
+  // cli.js: unhandledRejection and uncaughtException handlers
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

### What changed

**`engine/cli.js`** (lines 497-513): Added two process event handlers right after the existing SIGTERM/SIGINT handlers:

1. `process.on('unhandledRejection', ...)` — catches unhandled promise rejections
2. `process.on('uncaughtException', ...)` — catches synchronous throws that escape all try-catch blocks

Both handlers follow the same pattern:
- Extract stack trace (or string representation) from the error
- Log to stderr via `console.error` (immediate, unbuffered)
- Log via `shared.log('fatal', ...)` to engine/log.json (buffered)
- Call `shared.flushLogs()` to drain the buffer to disk
- Exit with code 1 so process monitors can detect the crash

Each step is individually try-caught so a failure in logging doesn't prevent the exit.

### Tests added

4 source-inspection tests in `test/unit.test.js`:
- Handler registration for both `unhandledRejection` and `uncaughtException`
- Each handler includes `log()`, `flushLogs`, and `process.exit(1)`

### Design decisions

- Used `shared.log` directly (not `e.log` via the engine() lazy-require) because at crash time the engine module may be in a broken state
- Wrapped logging in try-catch per the "do not silently swallow — log then exit" requirement — if logging fails, we still exit(1)
- Used `'fatal'` log level to distinguish crashes from normal `'error'` entries in log.json

yemi33 (commented):
VERDICT: APPROVE

Clean, minimal change that adds crash handlers to the engine process. 

**What it does:** Registers `unhandledRejection` and `uncaughtException` handlers in `cli.js` `start()`, right after the existing SIGTERM/SIGINT handlers. Both log via `console.error` + `shared.log()`, flush logs, and `process.exit(1)`.

**Review findings:**
- ✅ Correct placement after existing signal handlers
- ✅ Best-effort try-catch around `shared.log()` and `shared.flushLogs()` — appropriate since process state may be corrupted
- ✅ `shared.flushLogs` exists and is exported (engine/shared.js:63, :1105)
- ✅ Uses stack trace when available (`err.stack || err.message`)
- ✅ Tests follow the surrounding source-inspection pattern (consistent with codebase conventions)

**Minor non-blocking observation:** The tests are source-code string inspection rather than behavioral tests. This is a known tradeoff in this codebase (see team notes on "behavioral tests over source-code inspection"). For crash handlers specifically, behavioral testing would require spawning a child process and verifying exit code — reasonable to skip for this scope.

No issues found. Ship it.

yemi33:
Acknowledged Ripley's review — clean APPROVE with no blocking findings.

Re: non-blocking observation on source-code inspection tests — agreed this is a known tradeoff. Behavioral testing for crash handlers would require a child process spawn + exit code verification, which is out of scope for this change. The source-inspection pattern is consistent with surrounding tests (e.g., Bug #36 spawn-agent handler tests).

No changes needed.

— Dallas (Engineer)

PR #1020, opened 2026-04-14
fix(W-mnxw6zwk38g4): await discoverPipelineWork instead of fire-and-forget .catch

yemi33:
## Summary
- **What**: Fixed unhandled promise rejection risk in `discoverPipelineWork` call inside `discoverWork()` (engine.js ~line 2993)
- **Why**: The fire-and-forget `.catch(e => log(...))` pattern is unsafe — if `log()` throws inside the `.catch` handler, it becomes an unhandled rejection that (without a global handler) kills the process
- **Fix**: `await` the call inside the existing try-catch block, consistent with all other async phases in the tick loop (e.g., `pollPrStatus`, `processPendingRebases`, `pollPrHumanComments`)

## Files changed
- `engine.js` — replaced `discoverPipelineWork(config).catch(...)` with `await discoverPipelineWork(config)` (1 line)
- `test/unit.test.js` — updated existing test + added assertion that fire-and-forget `.catch()` is not used (1 consolidated test)

## Build & test
```bash
npm test    # 1678 passed, 1 pre-existing failure (SessionStart hook), 2 skipped
```

## Test plan
- [x] New test verifies `await discoverPipelineWork(config)` is present
- [x] New test verifies `.catch()` is NOT used in the pipeline block
- [x] All pre-existing tests still pass
- [x] No performance concern — function runs once per 60s tick, awaiting adds negligible latency

Built by Minions (Ralph — Engineer)

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -2990,7 +2990,7 @@ async function discoverWork(config) {
   // Pipeline orchestration — check stage completions and start ready stages
   try {
     const { discoverPipelineWork } = require('./engine/pipeline');
-    discoverPipelineWork(config).catch(e => log('warn', 'discover pipeline work: ' + e.message));
+    await discoverPipelineWork(config);
   } catch (e) { log('warn', 'discover pipeline work: ' + e.message); }
 
   // Periodic plan completion sweep — catch PRDs that completed while engine was down
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -11646,11 +11646,16 @@ async function testEngineAuditCritical() {
       'discoverPipelineWork must be async to use await');
   });
 
-  await test('engine.js handles async discoverPipelineWork', () => {
+  await test('engine.js awaits discoverPipelineWork inside try-catch (not fire-and-forget)', () => {
     const src = fs.readFileSync(path.join(MINIONS_DIR, 'engine.js'), 'utf8');
-    assert.ok(src.includes('discoverPipelineWork(config)'), 'must call discoverPipelineWork');
-    assert.ok(src.includes('.catch(') || src.includes('await discoverPipelineWork'),
-      'must handle the async result (await or .catch)');
+    const pipelineBlock = src.slice(
+      src.indexOf('Pipeline orchestration'),
+      src.indexOf('Periodic plan completion sweep')
+    );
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

**Root cause**: `discoverPipelineWork(config).catch(e => log('warn', ...))` at engine.js:2993 was fire-and-forget. The `.catch()` handler calls `log()`, which writes to `engine/log.json` via `mutateJsonFileLocked()`. If `log()` throws (e.g., disk full, lock contention), that error has no handler — it becomes an unhandled promise rejection.

**Fix**: Changed to `await discoverPipelineWork(config)` inside the existing try-catch. This is the same pattern used by every other async phase in `tickInner()` (pollPrStatus, processPendingRebases, pollPrHumanComments, reconcilePrs, etc.).

**Performance**: No impact. `discoverPipelineWork` doesn't contribute items to `allWork` — it only has side effects (checking stage completions, starting ready stages). The subsequent code doesn't benefit from parallelism. Tick runs every 60s, so even if pipeline discovery takes 100ms, it's negligible.

**TDD**: Wrote failing test first, then implemented the fix.

yemi33 (commented):
VERDICT: APPROVE

Clean fix. The fire-and-forget `discoverPipelineWork(config).catch(...)` pattern caused an unhandled rejection if the `.catch()` handler itself threw, and more importantly meant the outer `try-catch` block was bypassed entirely. Replacing with `await` routes all errors through the existing `try-catch`, which is the standard pattern used by every other phase in `discoverWork()`.

Test update is correct — asserts `await` is present and `.catch()` is absent in the pipeline block.

No issues found.

yemi33:
## Review Response

Reviewed the implementation notes and Lambert's approval review. No code changes requested — the fix is clean and correct.

**Verification:**
- ✅ Worktree clean, branch `work/W-mnxw6zwk38g4` up to date
- ✅ Code change is minimal and follows the existing pattern (all other phases in `discoverWork()` use `await` inside try-catch)
- ✅ Tests pass: **1678 passed**, 1 pre-existing failure (SessionStart hook — unrelated), 2 skipped
- ✅ The updated test correctly asserts both `await` presence and `.catch()` absence in the pipeline block

No additional changes needed.

PR #1021, opened 2026-04-14
fix(W-mnxw6zw9fuer): stop perpetual ADO poll retry when token unavailable

yemi33:
## Summary

- Removed `_adoPollHadAuthFailure = true` from the null-token early-return path in `pollPrStatus()` — this flag caused `needsAdoPollRetry()` to return true on every engine tick during `getAdoToken()`'s 10-minute backoff, triggering ~10 useless `pollPrStatus()` calls that just re-set the flag and returned early
- The flag is now only set for mid-poll auth errors (line 498 catch block) where a valid token expired during PR processing and immediate retry is warranted
- Added 2 regression tests: one verifies the null-token path does NOT set the flag, another verifies the flag is set exactly once in the function (in the catch block only)

## Test plan

- [x] New test: `ado.js pollPrStatus null-token early-return does NOT set _adoPollHadAuthFailure` — passes
- [x] New test: `ado.js pollPrStatus sets _adoPollHadAuthFailure only in auth error catch block` — passes
- [x] Existing test: `ado.js pollPrStatus resets _adoPollHadAuthFailure at start` — still passes
- [x] Existing test: `ado.js pollPrStatus sets _buildStatusStale on auth error` — still passes (the `_adoPollHadAuthFailure = true` in the catch block is preserved)
- [x] Full test suite: 1680 passed, 1 failed (pre-existing SessionStart hook issue), 2 skipped

### DIFF
--- a/engine/ado.js
+++ b/engine/ado.js
@@ -252,7 +252,8 @@ async function pollPrStatus(config) {
   const token = await getAdoToken();
   if (!token) {
     log('warn', 'Skipping PR status poll — no ADO token available');
-    _adoPollHadAuthFailure = true; // trigger retry on next tick
+    // Don't set _adoPollHadAuthFailure — getAdoToken() has its own 10-min backoff,
+    // and setting the flag would hammer pollPrStatus() every tick with no useful work.
     return;
   }
 
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -12411,6 +12411,28 @@ async function testPrDuplicateRaceFix() {
     assert.ok(resetIdx < fetchIdx, 'reset must happen before any adoFetch calls');
   });
 
+  await test('ado.js pollPrStatus null-token early-return does NOT set _adoPollHadAuthFailure', () => {
+    const src = fs.readFileSync(path.join(MINIONS_DIR, 'engine', 'ado.js'), 'utf8');
+    const pollFn = src.match(/async function pollPrStatus[\s\S]*?^}/m);
+    assert.ok(pollFn, 'pollPrStatus must exist');
+    // Extract the null-token early-return block
+    const tokenCheckMatch = pollFn[0].match(/if \(!token\)\s*\{[\s\S]*?\}/);
+    assert.ok(tokenCheckMatch, 'pollPrStatus must have a null-token check');
+    // The null-token early-return should NOT set the retry flag — backoff in getAdoToken handles timing
+    assert.ok(!tokenCheckMatch[0].includes('_adoPollHadAuthFailure = true'),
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

### Root Cause
In `engine/ado.js`, `pollPrStatus()` had `_adoPollHadAuthFailure = true` on line 255 in the null-token early-return path. This created a perpetual retry loop:

1. Auth fails → `getAdoToken()` sets 10-min backoff, returns `null`
2. `pollPrStatus()` resets flag (line 250) → `getAdoToken()` returns `null` (backoff) → sets flag (line 255) → returns early
3. `needsAdoPollRetry()` → `true` → engine calls `pollPrStatus()` on next tick
4. Repeat for entire 10-min backoff: ~10 useless calls, one per tick

### Fix
Removed `_adoPollHadAuthFailure = true` from the null-token path. The flag is now only set in the mid-poll catch block (line 498) for auth errors where a valid token expired during PR processing — the legitimate case for immediate retry.

### Files Changed
- `engine/ado.js` — removed flag-set from null-token path, added 2-line WHY comment
- `test/unit.test.js` — 2 new regression tests

### Verification
- `getAdoToken()` backoff is correct: 10-min backoff on failure (line 59), reset on success (`_adoTokenFailedUntil = 0`, line 52)
- Mid-poll auth error path (line 498) still sets the flag correctly for legitimate immediate-retry scenarios

yemi33 (commented):
VERDICT: APPROVE

## Review Summary

Clean, well-targeted fix. The bug: when `getAdoToken()` returns null, the old code set `_adoPollHadAuthFailure = true` on early return. Since `pollPrStatus` resets the flag to `false` at the top, this created a cycle where every tick re-entered `pollPrStatus` (via `needsAdoPollRetry()`), called `getAdoToken()` which returned null again (10-min backoff), logged a warning, set the flag again — perpetual no-op polling every 60s.

**Fix:** Simply don't set the flag on null-token early return. `getAdoToken()` already has its own 10-min backoff, so the retry flag adds no value here. The flag is correctly set only in the catch block for actual auth errors during API calls.

**Tests:** Two source-inspection tests verify the flag is absent from the null-token block and appears exactly once in the function (in the catch block). Tests pass. The 1 failure (`SessionStart hook uses http type`) is pre-existing and unrelated.

No issues found.

yemi33:
## Review Feedback Acknowledgment

Reviewed the human comment (Implementation Notes). This is a descriptive summary of the root cause and fix — no code changes are being requested.

**Review status:** Approved by Ripley with "No issues found."

**Tests:** 1680 passed, 1 failed (pre-existing `SessionStart hook uses http type` — unrelated), 2 skipped.

No changes needed. PR is ready to merge.

PR #1023, opened 2026-04-14
fix(W-mnxw6zwv7hh2): prevent _consolidationInFlight race from stale force-reset timeout

yemi33:
## Summary

- Store force-reset setTimeout handle in `_forceResetTimeout` and cancel it in `_clearProcessingState()` so it never fires against a subsequent consolidation
- Add `_cleared` idempotency guard to `_clearProcessingState()` so double-fire from error+close events on the same process doesn't clear a new consolidation's state
- Clear stale `_forceResetTimeout` at the top of `consolidateWithLLM()` to prevent timer leak from a prior run firing during a new one
- De-duplicated force-reset callback by delegating to `_clearProcessingState()` instead of inline cleanup

## Test plan

- [x] 6 new structural tests verify all race condition guards are in place
- [x] Full test suite: 1684 passed, 1 pre-existing failure (unrelated SessionStart hook test), 2 skipped

### DIFF
--- a/engine/consolidation.js
+++ b/engine/consolidation.js
@@ -18,6 +18,7 @@ const { getInboxFiles, getNotes, INBOX_DIR, ENGINE_DIR, MINIONS_DIR,
 // Track in-flight LLM consolidation to prevent concurrent runs
 let _consolidationInFlight = false;
 let _consolidationStartedAt = 0;
+let _forceResetTimeout = null; // force-reset handle; cancelled by _clearProcessingState
 const _processingFiles = new Set(); // files currently being consolidated (race guard)
 
 function consolidateInbox(config) {
@@ -113,6 +114,9 @@ Use today's date: ${dateStamp()}`;
 
 function consolidateWithLLM(items, existingNotes, files, config) {
 
+  // Cancel any stale force-reset from a prior run before starting fresh
+  clearTimeout(_forceResetTimeout);
+  _forceResetTimeout = null;
   _consolidationInFlight = true;
   _consolidationStartedAt = Date.now();
   for (const f of files) _processingFiles.add(f);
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation Notes

### Bug Analysis

The race condition in `engine/consolidation.js` had three manifestation paths:

1. **Primary (described in task):** The 3-minute timeout fires `killGracefully()` then sets a 10s `setTimeout` (force-reset). If the process then fires `error`, `_clearProcessingState()` clears `_consolidationInFlight`, but the anonymous 10s timeout is still pending. A new consolidation starts within 10s → the stale timeout fires and clears the new consolidation's flag → concurrent consolidation against the same inbox files.

2. **Double-fire:** Node.js `close` can fire after `error` for the same process (e.g., spawn failures). Without an idempotency guard, the second `_clearProcessingState()` call clears state belonging to a new consolidation that started between the two events.

3. **Cross-run timer leak:** If a previous consolidation's force-reset timeout is still pending when a new consolidation starts (e.g., 5-minute stale guard triggered), that stale timer fires mid-run and clears the new run's `_consolidationInFlight`.

### Changes

- `engine/consolidation.js`: 3 fixes + 1 simplification (force-reset delegates to `_clearProcessingState` instead of duplicating cleanup inline)
- `test/unit.test.js`: 6 structural tests covering all three race paths

### Files Changed
- `engine/consolidation.js` — the race condition fix
- `test/unit.test.js` — 6 new tests in `testConsolidationForceResetRace()`

yemi33 (commented):
VERDICT: APPROVE

## Summary

Clean, well-structured fix for the `_consolidationInFlight` race condition where a stale force-reset timeout from a prior timed-out consolidation could fire during a new run, incorrectly clearing the in-flight flag.

## What the fix does

1. **Stores the force-reset timeout handle** in module-level `_forceResetTimeout` so it can be cancelled
2. **Cancels stale timeouts on entry** to `consolidateWithLLM` — prevents prior run's timeout from interfering
3. **Cancels in `_clearProcessingState`** — normal completion cancels the force-reset before it fires
4. **Adds `_cleared` idempotency guard** — correct pattern for child processes where both `error` and `close` events can fire
5. **Resets `_consolidationStartedAt = 0`** in cleanup — prevents stale timestamp from confusing timeout checks
6. **Force-reset callback delegates to `_clearProcessingState()`** — single source of truth, no duplicated cleanup

## Tests

Six source-code inspection tests covering all fix aspects. Consistent with existing test patterns in the repo.

## No issues found

- Correct use of `clearTimeout` (safe to call with null)
- Force-reset log guarded by `!_cleared` — silent no-op on normal completion, logs only on actual force-reset
- No lock ordering concerns (no file locks involved)
- No cross-platform concerns

yemi33:
Reviewed the implementation notes — no code changes needed.

The comment from @yemi33 documents the three race condition manifestation paths and the fix approach. All align with the implementation:

1. **Primary race (stale force-reset fires during new consolidation)** → Fixed by storing handle in `_forceResetTimeout` and cancelling on entry to `consolidateWithLLM`
2. **Double-fire (error + close events)** → Fixed by `_cleared` idempotency guard in `_clearProcessingState`
3. **Cross-run timer leak** → Fixed by `clearTimeout(_forceResetTimeout)` in both `_clearProcessingState` and at the top of `consolidateWithLLM`

**Test results:** 1684 passed, 1 failed (pre-existing `SessionStart hook` issue, unrelated), 2 skipped. All 6 new `testConsolidationForceResetRace` tests pass.

No changes required — implementation notes acknowledged.

PR #1062, opened 2026-04-14
fix(W-mnywbse3jz13): cancel steering kill watcher on resume spawn

### LINKED ISSUE #1052: Steering kill watcher fires against resumed session, causing false error
## Bug

When steering kills an agent to resume with a human message, the 30s kill watcher fires against the newly-resumed session, killing it and marking the task as error — even though the resume succeeded and the agent already replied.

## Sequence (from engine log)

```
07:26:00Z  Steering: killing ripley (ripley-fix-mnyanfd3ptet) for session resume
07:26:00Z  Agent ripley exited with code 1  ← expected, kill succeeded
07:26:00Z  Steering: re-spawning ripley with --resume 1e4bbb57  ← resume launched
           [Ripley replies to human — resume successful]
07:26:31Z  Steering: ripley didn't exit 30s after kill — retrying  ← stale watcher fires
07:26:31Z  Steering resume for ripley exited with code 1  ← kills the resumed session
07:26:31Z  Completed dispatch: ripley-fix-mnyanfd3ptet (error)  ← false error
```

## Root Cause

The 30s kill watcher is started when the kill is issued. It doesn't get cancelled when the resume is successfully spawned, so it fires against the resumed process 30s later.

## Fix

Cancel the kill watcher once the `--resume` process has been successfully spawned (or once the agent produces any output confirming the resume succeeded).

yemi33:
## Summary

- **Bug**: The 30s kill watcher in `checkSteering` (timeout.js) was not cancelled when the `--resume` process was successfully spawned. The re-attach line in `engine.js:1055` set `_steeringAt: Date.now()` on the new tracking object, causing the watcher to fire 30s later — killing the resumed session and marking the task as error.
- **Fix**: Removed `_steeringAt` from the re-attached tracking object for the resumed process. The kill watcher only exists to handle cases where the original kill didn't take effect. Once the old process has exited and the resume is spawned, `_steeringAt` must not be present.
- **Tests**: Added 3 regression tests verifying that resumed process tracking does not carry `_steeringAt`, and that the kill watcher/skip guard logic remains correct.

Closes #1052

## Test plan

- [x] 3 new unit tests pass (`#1052` tag)
- [x] Full test suite: 1848 passed, 1 failed (pre-existing, unrelated)
- [ ] Manual: steer a running agent, confirm resumed session is NOT killed after 30s

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -1051,8 +1051,12 @@ async function spawnAgent(dispatchItem, config) {
         return;
       }
 
-      // Re-attach to existing tracking
-      activeProcesses.set(id, { proc: resumeProc, agentId, startedAt: procInfo.startedAt, sessionId: steerSessionId, _steeringAt: Date.now(), lastRealOutputAt: Date.now() });
+      // Re-attach to existing tracking — do NOT carry _steeringAt forward (#1052).
+      // The kill watcher in timeout.js fires 30s after _steeringAt is set. If we carry it
+      // into the resumed process, it kills the resumed session. The kill watcher only exists
+      // to handle cases where the original kill didn't take effect — once the process has
+      // exited and the resume is spawned, _steeringAt must not be present.
+      activeProcesses.set(id, { proc: resumeProc, agentId, startedAt: procInfo.startedAt, sessionId: steerSessionId, lastRealOutputAt: Date.now() });
 
       // Reset output buffers so post-completion parsing only sees the resumed session
       stdout = '';
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -7154,6 +7154,34 @@ async function testAgentSteering() {
       'spawn-agent should close stdin after writing prompt');
   });
 
+  // ── Steering kill watcher cancellation on resume (#1052) ───────────────────
+
+  await test('resumed process tracking does NOT carry _steeringAt (#1052)', () => {
+    // Bug: when the engine re-attaches after a successful steering resume spawn,
+    // it sets _steeringAt: Date.now() on the new tracking object. This causes
+    // checkSteering's 30s kill watcher to fire against the resumed session.
+    // Fix: the re-attached tracking must NOT have _steeringAt.
+    const resumeAttachLine = engineSrc.match(/activeProcesses\.set\(id,\s*\{[^}]*proc:\s*resumeProc[^}]*\}/);
+    assert.ok(resumeAttachLine, 'Should find activeProcesses.set for resumed process');
+    assert.ok(!resumeAttachLine[0].includes('_steeringAt'),
+      'Resumed process tracking must NOT include _steeringAt — kill watcher must not fire against resumed session');
+  });
+
+  await test('checkSteering kill watcher only fires when _steeringAt is set (#1052)', () => {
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

### Root Cause

In `engine.js:1055`, when re-attaching the resumed process to `activeProcesses`, the tracking object included `_steeringAt: Date.now()`. This is the flag that `checkSteering()` in `timeout.js:66` uses for the 30s kill watcher:

```js
if (info._steeringAt && Date.now() - info._steeringAt > STEERING_KILL_RETRY_MS) {
  shared.killImmediate(info.proc);  // ← kills the resumed session!
}
```

### Fix

One-line change: removed `_steeringAt` from the object passed to `activeProcesses.set()` for the resumed process. This is correct because:

1. **Kill watcher purpose**: Only exists to handle cases where `killImmediate` silently failed (e.g., orphaned subprocess). Once the old process has exited (we're in `onAgentClose`), this guard is no longer needed.
2. **Double-steering guard**: `_steeringAt` is also used in the skip guard (timeout.js:80). Without it on the resumed process, the agent can be steered again — which is the correct behavior.
3. **No side effects**: `_steeringMessage` and `_steeringSessionId` were already cleared at lines 983-984. Only `_steeringAt` was incorrectly carried forward.

### Tests Added

3 source-inspection tests following existing patterns:
- `resumed process tracking does NOT carry _steeringAt` — the key regression test
- `checkSteering kill watcher only fires when _steeringAt is set` — verifies guard
- `checkSteering skip guard prevents double-steering on resumed process` — verifies skip logic

yemi33:
## Review Response

Thanks for the detailed implementation notes — the analysis of the root cause and fix rationale is spot on.

### Verified:
- **Root cause confirmed**: `_steeringAt: Date.now()` in the resume re-attach line was indeed causing the 30s kill watcher to fire against the resumed session
- **Fix is minimal and correct**: Removing only `_steeringAt` from the re-attached tracking object is the right scope — `_steeringMessage` and `_steeringSessionId` were already cleared earlier (lines 983-984), so this completes the cleanup
- **Tests pass**: 1848 passed, 1 failed (pre-existing `SessionStart hook uses http type` — unrelated), 2 skipped

No changes needed — the implementation notes validate the fix rather than requesting modifications. The PR is ready for merge.

yemi33 (commented):
VERDICT: APPROVE

Clean, well-targeted fix. Removing `_steeringAt` from the resumed-process tracking object prevents the 30s kill watcher in `timeout.js:66` from firing against the freshly spawned resume session. The comment explaining *why* is excellent.

**What changed:**
- `engine.js:1054-1058` — dropped `_steeringAt: Date.now()` from the `activeProcesses.set()` call for resumed processes, added explanatory comment.
- `test/unit.test.js` — 3 new source-inspection tests confirming: (1) resumed tracking has no `_steeringAt`, (2) kill watcher guards on `_steeringAt`, (3) skip guard checks both flags.

**Verified:**
- `timeout.js` line 66 confirms kill watcher only fires when `info._steeringAt` is truthy — omitting it from the resume tracking is the correct fix.
- `timeout.js` line 80 skip guard (`_steeringMessage || info._steeringAt`) means a resumed process without `_steeringAt` is correctly steerable again.
- No other code path sets `_steeringAt` on resume — this was the only place.

No issues found.

PR #1140, opened 2026-04-15
feat(W-mo0kr8tuldlr): Intercept /loop skill invocations in CC and convert to watches

yemi33:
## Summary
- **CC system prompt**: Added explicit directive to NEVER use `/loop` for monitoring tasks, with trigger phrase examples directing CC to emit `create-watch` actions instead
- **Safety net in dashboard.js**: Added `_detectLoopInvocation()` that scans CC responses for /loop patterns and auto-synthesizes a `create-watch` action when no watch was already emitted — integrated in both streaming and non-streaming CC response paths
- **27 new tests**: Behavioral tests covering detection patterns, target extraction (PR numbers, work item IDs), interval parsing, condition inference (build-pass/build-fail/merged/completed/any), action shape validation, and integration point verification

## Test plan
- [x] All 27 new tests pass (`testLoopToWatchInterception`)
- [x] No regressions — baseline 2003 → 2030 passed, 1 pre-existing failure unchanged
- [ ] Manual: send CC a message like "keep an eye on PR 200 every 5 min" and verify a watch is created (not a /loop)
- [ ] Manual: verify if CC somehow emits /loop text, the safety net converts it to a watch entry in watches.json

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -607,6 +607,65 @@ function parseCCActions(text) {
   return { text: displayText, actions };
 }
 
+// ── /loop → create-watch safety net ──────────────────────────────────────────
+// CC sometimes invokes the /loop skill instead of emitting a create-watch action.
+// This pure function detects /loop invocation in CC response text and synthesizes
+// a create-watch action as a fallback. Returns null if no conversion needed.
+
+function _detectLoopInvocation(text, actions) {
+  if (!text) return null;
+  // If a create-watch action was already emitted, no fallback needed
+  if (actions && actions.some(a => a.type === 'create-watch')) return null;
+
+  // Check for /loop invocation patterns in CC response
+  const loopPatterns = [
+    /\/loop\b/i,
+    /\bloop skill\b/i,
+    /\bSkill.*\bloop\b/i,
+    /\bstarted.*\bloop\b/i,
+    /\bmonitoring.*\bloop\b/i,
+    /\binvok(?:e|ed|ing).*\bloop\b/i,
+  ];
+  if (!loopPatterns.some(p => p.test(text))) return null;
+
+  // Extract target — PR number or work item ID
+  const prMatch = text.match(/\bPR[- #]?(\d+)\b/i) || text.match(/\bpull[- ]request[- #]?(\d+)/i);
+  const wiMatch = text.match(/\bW-([a-z0-9]+)\b/);
+
+  let target = null, targetType = 'pr';
+  if (prMatch) {
+    target = prMatch[1];
+    targetType = 'pr';
+  } else if (wiMatch) {
+    target = 'W-' + wiMatch[1];
+    targetType = 'work-item';
+  }
+  if (!target) return null; // Can't synthesize without a target
+
+  // Extract interval (e.g. "every 15 minutes", "every 5m")
+  const intervalMatch = text.match(/every\s+(\d+)\s*(s|sec|seconds?|m|min|minutes?|h|hr|hours?)\b/i);
+  let interval = '5m';
+  if (intervalMatch) interval = intervalMatch[1] + intervalMatch[2][0];
+
+  // Infer condition from keywords
+  let condition = 'any';
+  if (/\bbuild\b/i.test(text) && /\b(?:pass(?:es|ing|ed)?|green|succeed(?:s|ed)?|success)\b/i.test(text)) condition = 'build-pass';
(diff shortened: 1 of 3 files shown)


yemi33:
## Implementation Notes

### Changes Made

**1. `prompts/cc-system.md`** — Added 3 lines near the `create-watch` documentation:
- Bold "NEVER use the /loop skill" directive explaining watches persist, /loop doesn't
- Trigger phrase examples: "keep an eye on X", "watch X every N min", "monitor X", "check X periodically", "ping me when X"
- New example: "keep an eye on PR 200 every 5 min" → create-watch with condition `any`

**2. `dashboard.js`** — Added `_detectLoopInvocation(text, actions)` pure function (lines 610-671):
- Skips if `create-watch` already in actions (primary prompt fix should work most of the time)
- Detects 6 patterns: `/loop`, `loop skill`, `Skill.*loop`, `started.*loop`, `monitoring.*loop`, `invok(e|ed|ing).*loop`
- Extracts target: PR number (`PR-123`, `PR 123`, `pull request 123`) or work item ID (`W-abc123`)
- Extracts interval: parses "every N min/h/s" patterns, defaults to "5m"
- Infers condition: build-pass, build-fail, merged, completed — defaults to "any"
- Returns null if no target found (can't create a watch without a target)
- Integrated at both `handleCommandCenter` (non-streaming, line 3585) and `handleCommandCenterStream` (streaming, line 3754) — both after `parseCCActions()` and before `executeCCActions()`

**3. `test/unit.test.js`** — 27 new behavioral + integration tests:
- Detection: all 6 patterns, empty/null text, normal text, already-has-watch skip
- Target extraction: PR formats, work item IDs, PR-over-WI precedence, no-target-returns-null
- Interval: "every N min", "every N h", default "5m"
- Condition inference: build-pass (green/passing/succeed), build-fail (failing/broken), merged, completed, default any
- Action shape: type, target, targetType, condition, interval, owner, description, stopAfter
- Integration: source-level checks that both CC paths call _detectLoopInvocation after parseCCActions

### Design Decisions
- **Pure function**: `_detectLoopInvocation` has zero side effects — it takes text+actions, returns an action or null. Easy to test, easy to reason about.
- **Prompt fix is primary, code is safety net**: The prompt directive should prevent most /loop usage. The code fallback only fires when the LLM ignores the prompt.
- **No target → no watch**: If we can't determine what to watch, we don't synthesize a garbage watch. Better to let it through than create a broken watch.

yemi33:
## Fix Review — No Actionable Issues Found

Reviewed the PR for issues to address. The only comment on this PR is implementation notes from @yemi33 — documenting the changes made, not requesting modifications.

### Verification
- **Tests**: 2030 passed, 1 pre-existing failure (SessionStart hook — unrelated), 2 skipped
- **Code review**: `_detectLoopInvocation` is a clean pure function with proper fallback behavior, integrated correctly at both CC response paths
- **Prompt changes**: Appropriate directive + examples in `cc-system.md`

No changes needed — implementation is solid as-is.

yemi33 (commented):
VERDICT: APPROVE

## Review: /loop → create-watch interception

**Clean, well-designed safety net.** The approach is sound — pure function for detection, hooked into both CC paths, comprehensive tests, and prompt-level guidance to prevent the problem at the source.

### Strengths

1. **Pure function design** — `_detectLoopInvocation` is side-effect-free and testable in isolation. Good separation from the integration points.
2. **Defense in depth** — prompt tells CC not to use /loop (preventive), detector catches it anyway (corrective). Both layers are needed.
3. **Correct skip logic** — returns `null` when `create-watch` already emitted, avoiding duplicate watches.
4. **Thorough test coverage** — 25+ behavioral tests covering detection, target extraction, interval parsing, condition inference, and action shape. Source-level integration checks verify both CC paths are wired up.

1. **False positive surface area** — Patterns like `/\bstarted.*\bloop\b/i` match generic "loop" usage (e.g., "started a loop to review PR 50 code"). Mitigated by the target requirement (no PR/WI → null), and CC responses rarely use "loop" generically alongside PR numbers. Acceptable tradeoff, but worth noting if false positives appear in practice.

2. **`stopAfter` semantics** — `condition === 'any' ? 0 : 1` is correct for absolute conditions (merged, build-pass, etc.) but slightly surprising for `completed`/`failed` which are also absolute. Works correctly since those are all non-`any`. ✓

3. **Duplicate logging** — Both `console.warn` and `shared.log('warn', ...)` fire on detection. The `shared.log` is wrapped in try-catch which is fine, but the dual logging means the same event appears in both stdout and the audit log. Minor — not worth changing.

No bugs, no security issues, patterns are consistent with the codebase. Ship it.

PR #1160, opened 2026-04-16
fix(W-mo1ircuy1r7h): scheduler enabled field falsy check treats undefined as disabled

yemi33:
## Summary

- **Bug:** `if (!sched.enabled) continue` in `engine/scheduler.js:117` treats `enabled: undefined` (field omitted) as disabled, silently skipping schedules that should run per the schema (`enabled?: boolean -- default true`).
- **Fix:** Changed to `if (sched.enabled === false) continue` — only schedules explicitly set to `false` are skipped; `undefined`/`null` default to enabled.
- **Tests:** Updated 3 source-inspection tests (Bug #35 and a duplicate guard test) to assert the corrected strict-false check instead of the buggy truthy check.

## Test plan
- [x] Updated existing tests to assert correct behavior (TDD: tests failed before fix, pass after)
- [x] Full test suite: 2119 passed, 0 failed

### DIFF
--- a/engine/scheduler.js
+++ b/engine/scheduler.js
@@ -114,7 +114,7 @@ function discoverScheduledWork(config) {
   mutateJsonFileLocked(SCHEDULE_RUNS_PATH, (runs) => {
     for (const sched of schedules) {
       if (!sched.id || !sched.cron || !sched.title) continue;
-      if (!sched.enabled) continue; // truthy check — matches dashboard UI badge behavior
+      if (sched.enabled === false) continue; // strict false check — undefined/null default to enabled per schema
 
       // Backward compat: runs[sched.id] can be a string (old format) or object (new format)
       const runEntry = runs[sched.id] || null;
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -11510,23 +11510,23 @@ async function testAuxModuleBugFixes() {
   });
 
   // Bug #35: scheduler treats undefined/null enabled as disabled
-  await test('scheduler.js: falsy enabled skips schedule', () => {
+  // Fix: only skip schedules explicitly set to enabled === false
+  // undefined/null should default to enabled (per schema: enabled?: boolean -- default true)
+  await test('scheduler.js: only sched.enabled === false skips schedule', () => {
     const src = fs.readFileSync(path.join(MINIONS_DIR, 'engine', 'scheduler.js'), 'utf8');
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

**Root cause:** `if (!sched.enabled) continue` is a truthy check that treats `undefined`, `null`, `0`, and `""` as falsy — so any schedule without an explicit `enabled: true` field was silently skipped.

**Fix:** `if (sched.enabled === false) continue` — strict equality means only `false` disables a schedule. `undefined` (field omitted), `null`, and `true` all pass through.

**Files changed:**
- `engine/scheduler.js:117` — the one-line fix
- `test/unit.test.js` — 3 tests updated from asserting `!sched.enabled` to asserting `sched.enabled === false`

**Testing:** TDD — updated tests first (they failed against buggy code), then applied the fix, confirmed 2119/2119 passing.

yemi33 (commented):
VERDICT: APPROVE

**Summary:** Single-line fix in `engine/scheduler.js` changes `!sched.enabled` → `sched.enabled === false`. Tests updated to match.

**Analysis:**

1. **Correctness:** The schema comment (line 14) documents `enabled?: boolean -- default true`. The old `!sched.enabled` check treated `undefined` as falsy, silently disabling schedules that omitted the field. `=== false` correctly defaults omitted/undefined to enabled, matching the documented schema.

2. **Edge cases covered:**
   - `enabled: true` → runs ✓
   - `enabled: false` → skipped ✓
   - `enabled: undefined` (omitted) → runs ✓ (was broken before)
   - `enabled: null` → runs ✓ (was broken before)

3. **Tests:** Three test blocks updated from asserting `!sched.enabled` to asserting `sched.enabled === false` and explicitly asserting `!sched.enabled` is NOT present. Tests are source-inspection style (consistent with surrounding tests), and the comments document expected behavior per case.

4. **Team notes context:** The `notes.md` entry from 2026-04-03 warned about `sched.enabled !== true` as a breaking change — that's a different (worse) fix. This PR correctly uses `=== false`, which is the right operator for "only skip when explicitly disabled."

No issues found. Clean, minimal, correct fix.

yemi33:
## Fix Response

Reviewed the feedback — the only comment on this PR is the original implementation notes (self-authored) and Ripley's approval review with `VERDICT: APPROVE` and no issues found.

No code changes needed. Build and tests remain passing (2119/0).

PR #1183, opened 2026-04-16
fix: sidecar oversized dispatch prompts to prevent dashboard OOM (#1167)

### LINKED ISSUE #1167: bug: fix-type prompts inline PR diff contexts into dispatch.json, causing OOM on startup
## Summary

Fix-type work items inline full PR diff contexts directly into their `prompt` fields in `dispatch.json`, causing file sizes to balloon to hundreds of MB per entry. This crashes the dashboard on startup when V8 tries to `JSON.parse()` ~1 GB.

## Root Cause

3 pending dispatch entries had PR diff contexts inlined directly into their prompt fields:
- ralph-fix: ~190 MB
- rebecca-fix: ~109 MB  
- lambert-fix: ~183 MB

Additionally, `cooldowns.json` was caching 505 MB of `pendingContexts`, bringing the total to ~1 GB that the dashboard tried to parse on startup — blowing past the V8 heap limit.

## Impact

- Dashboard fails to start (OOM / heap limit exceeded)
- Engine state unrecoverable without manual cleanup
- Fix tasks lost from queue (must be re-discovered on next tick)

## What Was Cleaned (manually)

- `dispatch.json`: 491 MB → 5.6 MB (removed 3 bloated pending fix entries)
- `cooldowns.json`: 509 MB → 2 bytes (cleared stale cooldown entries with cached PR contexts)

## Suggested Fix

1. **Never inline PR diff / context into `prompt` fields** stored in `dispatch.json`. Write large context blobs to a sidecar file (e.g. `engine/contexts/<id>.txt`) and reference it by path. The agent reads the file at spawn time.
2. **Cap or evict `cooldowns.json` `pendingContexts`** — stale entries with large payloads should not persist indefinitely.
3. **Add a prompt size guard** in the dispatch mutation path (`mutateDispatch` / `mutateJsonFileLocked` in `engine/shared.js`): if a prompt exceeds a threshold (e.g. 1 MB), write it to a sidecar instead of inlining, with a warning logged.
4. **Add a startup size check** in `dashboard.js` / `engine.js` before `JSON.parse()` on large state files, with a clear error message pointing to the bloated file rather than a silent OOM crash.

## Files Involved

- `engine/dispatch.json` — prompt field bloat
- `engine/cooldowns.json` — pendingContexts bloat  
- `engine/shared.js` — dispatch mutation path (where size guard should be added)
- `engine/spawn-agent.js` — where large context could be written to sidecar instead

yemi33:
Closes #1167.

## Problem

Fix-type work items inlined full PR diff / build-log / coalesced-human-feedback contexts directly into the `prompt` field of their dispatch entries. Three pending entries grew to 190 MB + 109 MB + 183 MB, and `cooldowns.json` cached another 505 MB of `pendingContexts`. On dashboard startup, `JSON.parse(~1 GB)` blew past the V8 heap limit and the dashboard would not come up at all.

## Fix

Four independent layers, each of which is sufficient to prevent a recurrence, together making the failure mode impossible:

1. **Sidecar oversized prompts.** `engine/shared.js` now exposes `sidecarDispatchPrompt`, `resolveDispatchPrompt`, `deleteDispatchPromptSidecar`, `dispatchPromptSidecarPath`. Prompts larger than `ENGINE_DEFAULTS.maxDispatchPromptBytes` (1 MB) are written to `engine/contexts/<id>.md`; `dispatch.json` only keeps a short stub naming the sidecar path plus `_promptFile` / `_promptBytes` markers. The contexts dir is gitignored.
2. **Automatic size guard in `mutateDispatch`.** `_sidecarOversizedPrompts` runs on every write, so a single new item cannot bloat the queue even if the discovery path forgets to sidecar. `completeDispatch` deletes the sidecar to keep `engine/contexts/` bounded.
3. **`spawnAgent` reads via `resolveDispatchPrompt`.** Agents still see the full prompt — the sidecar is a storage detail, not an agent-visible one.
4. **Per-entry byte cap on cooldown `pendingContexts`.** Count-capping at 20 entries was insufficient: 20 × 50 MB is still 1 GB. Each string field is individually truncated to `maxPendingContextEntryBytes` (256 KB) with a visible `[truncated: ... exceeded X KB]` marker.
5. **Startup guard.** `assertStateFileSize` is called on `dispatch.json` and `cooldowns.json` from `dashboard.js` and `engine/cli.js start`; when a state file exceeds 100 MB the process exits with code 78 and an actionable message naming the bloated file and pointing at `engine/contexts/`, instead of silently OOMing.

## Tests

11 new tests cover: sidecar small-prompt no-op, sidecar oversized-prompt behavior (file written, stub replaces inline, full content restored via `resolveDispatchPrompt`), sidecar cleanup on `completeDispatch`, `mutateDispatch` auto-sidecar on insert (asserts `dispatch.json` size stays < 100 KB), `assertStateFileSize` throw/no-throw cases, dashboard + engine startup wiring, and cooldown entry truncation. Full suite: **2211 pass / 0 fail / 3 skip**.

## Test plan

- [x] Run `npm test` — all pass.
- [ ] On a machine with a pre-existing bloated `dispatch.json`, `minions start` / `minions dash` now prints a clear abort message instead of OOMing.
- [ ] Dispatch a fix-type task whose review note is large; confirm a file appears under `engine/contexts/`, `dispatch.json` stays small, and the agent still receives the full prompt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ engine/kb-swept.json
 engine/watches.json
 # All transient temp files (prompts, pids, logs) go in engine/tmp/
 engine/tmp/
+# Large-prompt sidecar files referenced by dispatch.json entries (#1167)
+engine/contexts/
 # Backup/lock files from safeWrite atomic writes
 engine/*.bak
 engine/*.backup
--- a/dashboard.js
+++ b/dashboard.js
@@ -32,6 +32,24 @@ const { getAgents, getAgentDetail, getPrdInfo, getWorkItems, getDispatchQueue,
   getEngineLog, getMetrics, getKnowledgeBaseEntries, timeSince,
   MINIONS_DIR, AGENTS_DIR, ENGINE_DIR, INBOX_DIR, DISPATCH_PATH, PRD_DIR } = queries;
 
+// Startup size guard (#1167): fail fast with a clear error when dispatch.json /
+// cooldowns.json have ballooned past ENGINE_DEFAULTS.maxStateFileBytes. Without
+// this, V8 silently OOMs on JSON.parse(~1 GB) and the operator has no hint as to
+// which file is bloated. The thrown error names the file and directs to
+// engine/contexts/ where sidecars live.
+(() => {
+  const stateFiles = [
+    DISPATCH_PATH,
+    path.join(ENGINE_DIR, 'cooldowns.json'),
+  ];
+  for (const fp of stateFiles) {
+    try { shared.assertStateFileSize(fp); } catch (e) {
+      console.error('\n[dashboard] STARTUP ABORTED — ' + e.message + '\n');
+      process.exit(78); // 78 = configuration error; consistent with spawn-agent.js
+    }
+  }
+})();
+
 const PORT = parseInt(process.env.PORT || process.argv[2]) || 7331;
 let CONFIG = queries.getConfig();
 let PROJECTS = _getProjects(CONFIG);
--- a/engine.js
+++ b/engine.js
@@ -355,7 +355,11 @@ async function recoverPartialWorktree(rootDir, worktreePath, branchName, gitOpts
 }
 
 async function spawnAgent(dispatchItem, config) {
-  const { id, agent: agentId, prompt: taskPrompt, type, meta } = dispatchItem;
+  const { id, agent: agentId, type, meta } = dispatchItem;
+  // Resolve prompt — prefers sidecar file when dispatchItem._promptFile is set
+  // (large prompts are written to engine/contexts/<id>.md to keep dispatch.json
+  // small — see shared.sidecarDispatchPrompt / #1167).
+  const taskPrompt = shared.resolveDispatchPrompt(dispatchItem);
   const claudeConfig = config.claude || {};
   const engineConfig = config.engine || {};
   const startedAt = ts();
--- a/engine/cli.js
+++ b/engine/cli.js
@@ -51,6 +51,18 @@ function handleCommand(cmd, args) {
 
 const commands = {
   start() {
+    // Startup state-file size guard (#1167): dispatch.json / cooldowns.json
+    // bloated past 100 MB silently OOMed V8 on JSON.parse at startup. Fail fast
+    // with an actionable message instead.
+    try {
+      for (const fp of [DISPATCH_PATH, path.join(ENGINE_DIR, 'cooldowns.json')]) {
+        shared.assertStateFileSize(fp);
+      }
+    } catch (stateErr) {
+      console.error('\n[engine] STARTUP ABORTED — ' + stateErr.message + '\n');
+      process.exit(78); // 78 = configuration error
+    }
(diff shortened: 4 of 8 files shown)


yemi33 (commented):
VERDICT: APPROVE

Reviewed the full diff. This is a well-structured, layered fix for the OOM-on-startup bug (#1167):

**Strengths**
- Five independent defenses stack cleanly: (1) sidecar oversized dispatch prompts, (2) auto-sidecar inside `mutateDispatch` so no caller can bypass it, (3) cooldown per-entry truncation on top of the existing count cap, (4) prompt + sidecar cleanup in `completeDispatch`, (5) startup size guard on both `dashboard.js` and `engine/cli.js start` with exit code 78 and a message that names the bloated file. If any one layer regresses, the others still prevent the GB-scale bloat.
- `spawnAgent` correctly reads via `resolveDispatchPrompt(dispatchItem)` (engine.js:358), so the agent always sees the full prompt even though dispatch.json only holds the stub.
- `PROMPT_CONTEXTS_DIR` added to `.gitignore`; sidecar filenames sanitized (`[^a-zA-Z0-9._-]` → `-`) before filesystem write — no path traversal.
- `_promptContextsDir()` lazy resolution lets tests override `MINIONS_DIR`; cache-busting pattern in the test suite is consistent.
- Test coverage is thorough: small-prompt no-op, oversize sidecar round-trip, resolve returning full content, cleanup on `completeDispatch`, auto-sidecar via `mutateDispatch`, `assertStateFileSize` throw/no-throw/missing-file cases, plus string-match guards that dashboard.js / cli.js / engine.js actually call the helpers.
- Constants live in `ENGINE_DEFAULTS` (no magic numbers), sizes documented inline with rationale.

**Non-blocking observations**
1. `_sidecarOversizedPrompts` performs `safeWrite()` of the sidecar *inside* the `mutateJsonFileLocked(DISPATCH_PATH, ...)` callback (engine/dispatch.js:40-51). Per the lock-ordering rule in CLAUDE.md (\"Keep lock callbacks fast\"), disk I/O for a 2–25 MB sidecar while holding the dispatch lock will block concurrent consumers for tens to hundreds of ms. Practically fine today because dispatch mutations are serialized anyway, but worth noting if contention shows up. Writing the sidecar *before* entering the lock and only swapping refs inside would be the stricter pattern.
2. If a caller later shrinks `item.prompt` back below threshold *without* going through `completeDispatch`, the stale `_promptFile` pointer persists and `resolveDispatchPrompt` will still read the (now-stale) sidecar. No current code path does this, but a defensive clear in `sidecarDispatchPrompt` when `byteLen <= threshold && item._promptFile` would future-proof it.
3. The exported `PROMPT_CONTEXTS_DIR` constant captures the value at shared.js load time, while helpers use `_promptContextsDir()` lazily. The comment acknowledges this; consider either dropping the constant or exporting the lazy resolver to avoid confusion.

None of these block merge. Ship it.

PR #1198, opened 2026-04-16
fix(W-mo248lkjwgsu): stamp live-output.log stub before spawn

yemi33:
## Summary

Writes a stub to `live-output.log` **before** `runFile` spawns the child
process so the orphan detector can distinguish three failure modes that
previously looked identical:

| Log-file state at orphan check | Meaning |
|---|---|
| no file | spawn call itself threw before the stub write |
| stub only (~180 B) | process started but died before its first write |
| stub + content (>500 B) | genuine hang — only case that warrants orphan kill+retry |

Orphan/hung warn lines in `engine/timeout.js` now end with
`[logExists=<bool> logSize=<bytes>]` so the audit trail captures which
failure mode occurred.

- `engine.js` — moved the `liveOutputPath` stamp + rotation block to
  before `runFile`; wrapped `runFile` in try/catch so a synchronous
  spawn failure appends `[process-exit] spawn-failed` before rethrowing
  (dispatch loop at `engine.js:3525` already handles `spawnAgent`
  throws).
- `engine/timeout.js` — added `_logState` annotation to both the
  `Orphan detected` and `Hung agent` warn lines.
- `test/unit.test.js` — TDD tests that failed before implementation and
  pass after.

## Test plan

- [x] `node test/unit.test.js` — 2316 passed / 0 failed / 3 skipped
- [x] `node -e "require('./engine.js'); require('./engine/timeout.js')"` — modules load without error
- [x] New tests fail on the pre-change source (confirmed before
  implementation)
- [x] Stub-write error path is guarded (try/catch around `safeWrite`) so
  a failing log stamp cannot block agent spawn

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -871,25 +871,16 @@ async function spawnAgent(dispatchItem, config) {
   const spawnScript = path.join(ENGINE_DIR, 'spawn-agent.js');
   const spawnArgs = [spawnScript, promptPath, sysPromptPath, ...args];
 
-  const proc = runFile(process.execPath, spawnArgs, {
-    cwd,
-    stdio: ['pipe', 'pipe', 'pipe'],
-    env: childEnv,
-  });
-
-  const MAX_OUTPUT = 1024 * 1024; // 1MB
-  let stdout = '';
-  let stderr = '';
-  let lastOutputAt = Date.now();
-  let heartbeatTimer = null;
-  let _trustCheckDone = false;
-  const _spawnTime = Date.now();
-
-  // Live output file — written as data arrives so dashboard can tail it
+  // Live output file — stamped BEFORE child process is spawned (#W-mo248lkjwgsu).
+  // Writing the stub pre-spawn lets the orphan detector distinguish three failure modes
+  // that otherwise look identical:
+  //   1. No log file          → spawn call itself threw before reaching this line
+  //   2. Log has stub only    → process started but died before its first write
+  //   3. Log has stub + ...   → process alive but hung (the only case that warrants orphan kill+retry)
   const liveOutputPath = path.join(AGENTS_DIR, agentId, 'live-output.log');
 
   // Rotate previous live output to preserve session history (fixes #543: orphan recovery overwrites)
-  // Only rotate if the existing file has meaningful content (beyond just the header)
+  // Only rotate if the existing file has meaningful content (beyond just the header stub)
   const LIVE_OUTPUT_SPARSE_THRESHOLD = 500; // bytes — header + init JSON is typically < 500
   try {
     if (fs.existsSync(liveOutputPath)) {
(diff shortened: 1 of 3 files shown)


yemi33:
## Implementation notes

**Why reorder matters.** `runFile()` (thin wrapper over `child_process.spawn`) can throw synchronously on malformed args or ENOENT on the executable, in addition to emitting the `'error'` event async. Stamping the log *before* the spawn gives us a reliable "the engine tried to spawn this agent" marker regardless of which failure path hits.

**Rotation guard is byte-based** (`LIVE_OUTPUT_SPARSE_THRESHOLD = 500`). A stub-only log (~180 bytes from `# Live output for... # Started:... # Task:... [iso] spawn: agent=X item=Y`) does **not** trigger rotation, so a genuine previous-session log in `live-output-prev.log` survives repeated spawn failures without getting overwritten by a string of stub files.

**Dispatch-queue safety.** The pending→active mutation in `spawnAgent` happens at `engine.js:1325`, which is *after* the spawn call succeeds. If our rethrow fires (sync spawn failure), the item is still in `pending` — no cleanup is needed on the rethrow path. The dispatch loop at `engine.js:3525-3528` catches `spawnAgent` throws and re-queues via `mutateWorkItems` when needed.

**The try/catch around `safeWrite` for the stub is intentional.** If the filesystem rejects the stamp write (disk full, permission issue), we log at `warn` and still attempt the spawn — log-stamping failure must not block agent dispatch. Conversely, the try/catch around `runFile` *rethrows* after appending the diagnostic line, because spawn failure *is* a reason to fail the dispatch.

**TDD**: both new tests were confirmed failing on the pre-change source before implementing, then pass after.

yemi33 (commented):
VERDICT: APPROVE

Reviewed diff (engine.js, engine/timeout.js, test/unit.test.js) end-to-end. The change is sound and the failure-mode taxonomy in the commit message maps cleanly onto the implementation.

## What works

- **Stub-before-spawn ordering is correct.** `safeWrite(liveOutputPath, ...)` is now unambiguously before `runFile(...)` in the source, and the `[iso] spawn: agent=X item=Y` marker is enough to distinguish case 2 (startup crash) from case 3 (hang) by file size alone (stub ≈180 B vs `LIVE_OUTPUT_SPARSE_THRESHOLD=500` guard, so rotation never clobbers the previous real session). (source: `engine.js:880-902`)
- **Defensive posture is right.** The stub-write is wrapped so a filesystem failure while stamping cannot block agent dispatch (logged at `warn`), while the `runFile` wrap *rethrows* after appending the diagnostic marker — spawn failure is a real dispatch failure and the dispatch loop at `engine.js:3525-3528` already handles `spawnAgent` throws correctly. (source: `engine.js:898-917`, `engine.js:3525`)
- **Dispatch-state invariant preserved.** The pending→active mutation happens at `engine.js:1325-1337`, *after* the spawn call succeeds, so a synchronous spawn throw leaves the item in `pending`. No stale `active` entry, no cleanup needed on the rethrow path. Verified.
- **Orphan annotation is correctly placed.** `_logState` is captured once per iteration and appended to both the `Orphan detected` and `Hung agent` warn lines. `liveLogPath` is reused from `timeout.js:165`, no redundant path construction.
- **TDD order followed.** Tests added first (confirmed failing on pre-change source per the author's description), pass after. Full suite: 2316 / 0 / 3.
- **`[process-exit] spawn-failed` marker can't leak.** `safeWrite` on the next spawn attempt does `writeFileSync(tmp) + rename`, which replaces the file atomically — the stale marker is overwritten on retry. And `checkTimeouts` only iterates `dispatch.active` (source: `engine/timeout.js:157`), so a pending-queued item with the marker would never be scanned.

## Non-blocking observations

- **`fs.statSync` runs every tick for every active dispatch, even healthy ones.** `_logState` is computed at `timeout.js:316-320` before the `if (!hasProcess && silentMs > effectiveTimeout ...)` gate, so it costs one syscall per active agent per tick regardless of whether anyone is timing out. Could be moved inside the two failure branches to save ~N syscalls/minute. Negligible at current scale (N is typically ≤ `maxConcurrent=5`), so not worth blocking on — just flagging for the next pass. (source: `engine/timeout.js:316-320`)
- **Tests are source-code-inspection style.** Consistent with the existing live-output tests at `test/unit.test.js:11990-12011`, so this is the local convention for spawn-chain testing (hard to exercise the real spawn path in unit tests without heavy mocking). The team notes from 2026-04-03 call out a general preference for behavioral tests (`knowledge/reviews/2026-04-03-dallas-review-feedback-for-dallas.md`), but the spawn path is one of the reasonable exceptions and the string-matched assertions here (`safeWrite(liveOutputPath`, `runFile(process.execPath, spawnArgs`, ordering check, `item=${id}` marker) are specific enough to fail on real regressions.

PR #1200, opened 2026-04-17
feat(W-mo25loq8kjer): seed realActivityMap at spawn time, stamp pid in live-output

yemi33:
## Summary

- Seed `realActivityMap` and stamp a `pid` line to `live-output.log` **immediately after `runFile()` returns** in `spawnAgent()`, before any handlers or timers register.
- Remove the duplicate late-init seed 380 lines below — it could clobber real stdout timestamps that arrived during the synchronous setup window.
- Update `engine/timeout.js` orphan diagnostic (`_logState`) to surface `pidPresent=true|false`, reading only the first 4KB of the log.

## Why

`timeout.js` uses `realActivityMap.get(item.id)` as last-activity for tracked processes and falls back to `item.started_at` (dispatch time) when the map is empty.

For write tasks, dispatch → spawn can span 20-60s of worktree setup and dependency branch merging. For read-only tasks (explore, security audit, large scans) the process may then produce no stdout for several minutes. Combined, the heartbeat clock was eating a big chunk of `heartbeatTimeout` *before the process had emitted a single byte* — causing premature hung-agent kills on legitimate long-running agents.

Seeding at actual spawn time anchors the clock correctly. The pid line gives orphan detection a deterministic signal to tell "spawn died before first write" (stub only) from "process started and is hung" (stub + pid line).

## Subtleties

- **Seed BEFORE registering handlers.** `proc.on('error', ...)` calls `realActivityMap.delete(id)` on synchronous spawn failures. Seeding first makes the set→delete sequence symmetric — otherwise an immediate spawn error (fires via `process.nextTick`) could delete a not-yet-set entry.
- **Why the old late-init was wrong.** stdout/stderr handlers were registered in the window between spawn and the old re-seed. Although JS single-threading prevents stdout from firing during synchronous code, keeping the old `realActivityMap.set` at line 1297 would unconditionally overwrite the real-activity timestamp with a later `Date.now()` — a lie if real output had already arrived during the synchronous setup.
- **`proc.pid ?? 'unknown'`.** Defensive fallback for the rare case where pid isn't yet assigned.

## Test plan

- [x] TDD: wrote 4 failing tests first, then implemented (under `testIssue716HeartbeatFeedbackLoop`)
  - Seed happens between `runFile(process.execPath, spawnArgs` and `proc.stdout.on` in source order
  - `fs.appendFileSync(liveOutputPath, ...pid:...)` appears in the same post-spawn window
  - No duplicate `Initialize real activity at spawn time` comment survives (guards against silent revert)
  - `timeout.js` `_logState` includes `pidPresent` annotation
- [x] Full unit suite: **2320 passed, 0 failed, 3 skipped**
- [ ] Verify in live engine: read-only explore task survives past old premature-kill window

## Files changed

| File | Delta |
|------|-------|
| `engine.js` | +23 / -2 — post-spawn seed + pid line, removed duplicate late-init |
| `engine/timeout.js` | +25 / -7 — `pidPresent` annotation, 4KB head read |
| `test/unit.test.js` | +52 — 4 new tests |

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -916,6 +916,23 @@ async function spawnAgent(dispatchItem, config) {
     throw spawnErr;
   }
 
+  // Seed realActivityMap and stamp PID immediately — BEFORE any handlers or timers (#W-mo25loq8kjer).
+  // Why NOW, not later in the function:
+  //  1. Heartbeat clock anchoring. timeout.js uses realActivityMap as the last-activity timestamp for
+  //     tracked processes; when the map has no entry, it falls back to item.started_at (dispatch time,
+  //     which is 20-60s before actual spawn for write tasks doing worktree setup). Read-only tasks
+  //     that produce no stdout for minutes (explore, security audit, large scans) were hitting
+  //     heartbeatTimeout prematurely — clock had already been running since dispatch.
+  //  2. Error-handler race. The `proc.on('error', ...)` handler below calls realActivityMap.delete(id)
+  //     on synchronous spawn failures. Seeding before registering handlers ensures delete sees a value
+  //     to clear rather than leaving an absent-then-absent no-op that downstream code must guard.
+  //  3. Orphan diagnostics. The PID line gives timeout.js a deterministic way to tell "spawn died
+  //     before first write" (stub-only log) from "process started and is hung" (stub + pid line).
+  realActivityMap.set(id, Date.now());
+  try {
+    fs.appendFileSync(liveOutputPath, `[${new Date().toISOString()}] pid: ${proc.pid ?? 'unknown'}\n`);
+  } catch { /* log stamp is best-effort — don't block spawn on fs failure */ }
+
   const MAX_OUTPUT = 1024 * 1024; // 1MB
   let stdout = '';
   let stderr = '';
@@ -1292,9 +1309,11 @@ async function spawnAgent(dispatchItem, config) {
     }
   }, 3000);
 
-  // Track process — even if PID isn't available yet (async on Windows)
+  // Track process — even if PID isn't available yet (async on Windows).
+  // realActivityMap was already seeded immediately after runFile() returned (#W-mo25loq8kjer);
+  // don't re-seed here — the stdout/stderr handlers above can already have updated it with
+  // a fresher timestamp, and overwriting would clobber the real "last activity" signal.
   activeProcesses.set(id, { proc, agentId, startedAt, sessionId: cachedSessionId });
-  realActivityMap.set(id, Date.now()); // Initialize real activity at spawn time
 
   updateAgentStatus(id, AGENT_STATUS.RUNNING, `Process spawned for ${agentId}`);
 
--- a/engine/timeout.js
+++ b/engine/timeout.js
@@ -308,15 +308,33 @@ function checkTimeouts(config) {
     const procInfo = activeProcesses.get(item.id);
     if (procInfo?._steeringAt && Date.now() - procInfo._steeringAt < 60000) continue;
 
-    // Capture live-output.log file state for orphan/hung diagnostics (#W-mo248lkjwgsu).
-    // Three distinguishable failure modes:
-    //   logExists=false           → spawn call itself threw, no log ever written
-    //   logExists=true, size~stub → process died at startup (stub-only content)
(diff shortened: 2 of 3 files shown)


yemi33:
## Implementation notes

Three-part surgical fix — didn't touch any behavior outside `spawnAgent()` post-spawn setup and the `_logState` diagnostic string.

### Ordering rationale (this is the part that matters)

The fix hinges on when we seed `realActivityMap`, not what we write to it. Before this change, the seed was at the very end of `spawnAgent()` (~line 1297), after 380 lines of handler registration and timer setup. Although that's still synchronous, two subtle things broke:

1. **Error-handler race.** `proc.on('error', ...)` — registered at line 1280 — calls `realActivityMap.delete(id)` on spawn failures. Node can fire 'error' via `process.nextTick` after the `spawn()` call returns. If the error fired before line 1297, the delete hit a not-yet-set entry, then the set at 1297 leaked a ghost entry the map never cleaned up.
2. **Clock anchoring.** `timeout.js:172` only reads `realActivityMap.get(item.id)` if `realActivityMap.has(item.id)`. For the entire dispatch → spawn window (20-60s for write tasks with worktree setup + dep branch fetch), the map has no entry → timeout.js falls back to `item.started_at` (dispatch time). Combined with read-only tasks that produce no stdout for several minutes, the agent looked "silent for `heartbeatTimeout`" long before it had started doing real work.

Seeding **immediately after `runFile()` returns** — before any handler or timer registration — fixes both.

### The pid line

`[<iso>] pid: <N>` to `live-output.log`, written synchronously after the spawn returns. Paired with `_logState` in `timeout.js`, the orphan detector now distinguishes four states:
- `logExists=false` → spawn call itself threw
- `logExists=true pidPresent=false` → stub was written but spawn died before pid stamp (rare)
- `logExists=true pidPresent=true` with stub-only size → process spawned but never produced stdout (true "hang at startup")
- `logExists=true pidPresent=true` with size > pid line → genuine mid-run hang

The 4KB head read in `timeout.js:329` is deliberate — pid line is stamped right after the stub, so we never need to touch the tail for this annotation.

### What I did NOT change

- `proc.on('error')` / `proc.on('close')` handlers stay where they are. They already call `realActivityMap.delete(id)` correctly.
- The heartbeat `setInterval` still writes `[heartbeat]` lines to live-output.log. That wasn't the bug — the bug was the timeout check using those writes as activity signals. The existing `realActivityMap` design already solves that; we just weren't seeding the map early enough.
- No test file moves. Added 4 new tests inside the existing `testIssue716HeartbeatFeedbackLoop` suite since this is directly related.

### Verification

`npm test` → **2320 passed, 0 failed, 3 skipped** (no regressions).

yemi33 (commented):
VERDICT: APPROVE

Clean fix. The seeding moves from post-handler-registration to immediately after `runFile()` returns, which correctly anchors the heartbeat clock to actual spawn time instead of dispatch time (which can be 20-60s earlier due to worktree setup). This matters most for silent read-only tasks (explore, security audits).

Verified:
- Seed is after the spawn try/catch `throw spawnErr`, so failed spawns don't leave a stale entry (engine.js ~916).
- Old late-init at ~1313 removed; the comment correctly notes stdout/stderr handlers may have already written a fresher timestamp that we shouldn't clobber.
- `pid:` line format `[<iso>] pid: <N>` is grep-stable and timeout.js regex `\]\s+pid:\s+\d+` matches it.
- Head-only 4KB read in timeout.js is a good perf choice; file descriptor is closed inside the inner try, and the outer catch swallows read failure to keep `pidPresent=false` — acceptable.
- Tests are source-string-matching style (consistent with this file's convention for structural invariants) and assert on both the positional seed and the removal of the old seed.

Non-blocking nits:
- The inner `fs.openSync`/`readSync`/`closeSync` block in timeout.js will leak an fd if `readSync` throws (closeSync never runs). Since the outer catch just shrugs, it's rare and self-healing, but a try/finally around closeSync would be cleaner.
- Consider also guarding `appendFileSync` on the pid line with the same freshness: if `proc.pid` is `undefined` (Windows async), the line writes `pid: unknown` — harmless but defeats the diagnostic. A tiny follow-up could defer that write until `proc.pid` is known, or write it from the stdout 'spawn' event.

Issue #1204, opened 2026-04-17
Pending items pre-assigned to unspawned temp agents never get reassigned to idle named agents


yemi33:
## Bug

When the engine dispatches a batch of fix/review items (e.g. after a PR poll sweep), it pre-assigns agent IDs at creation time — including temp agent IDs. These temp agents don't exist yet; they're only spawned when the item actually dispatches.

If `maxConcurrent` is already saturated when those items are created, the items sit in the pending queue with temp agent IDs. The reassignment logic at the dispatch loop checks `busyAgents.has(item.agent)` — but unspawned temp agents are never in `busyAgents`, so the condition never fires. The items wait indefinitely, even if named agents (dallas, lambert, rebecca, etc.) are completely idle.

## Root Cause

`engine.js` reassignment threshold (`agentBusyReassignMs`) only triggers when `busyAgents.has(item.agent)` is true. Unspawned temp agents are not in `busyAgents`, so the 10-minute reassignment window never starts.

## Steps to Reproduce

1. Have `maxConcurrent: 6` with all 6 slots active
2. Trigger a large batch dispatch (e.g. ADO poll detecting many build failures)
3. Items are created pre-assigned to temp agents and sit in pending
4. Named agents are idle but receive no work — queue appears stuck

## Expected Behavior

If a pending item is assigned to a temp agent that is not currently active/running, the engine should treat it as unassigned immediately and let `resolveAgent` pick a free named agent on the next tick.

## Fix

In the pending dispatch loop, before the `busyAgents` check, detect if `item.agent` is a temp agent not present in the active set — and if so, clear the agent assignment so `resolveAgent` can route it to an available named agent.

```js
// If assigned to a temp agent that was never spawned, treat as unassigned
const isUnspawnedTemp = item.agent?.startsWith('temp-') && !activeAgentIds.has(item.agent);
if (isUnspawnedTemp) {
  delete item.agent;
  delete item.agentName;
  delete item.agentRole;
}
```

## Workaround

Manually clear `agent`/`agentName`/`agentRole` from stuck pending items in `dispatch.json` for temp agents not in the active set.

PR #1210, opened 2026-04-17
fix: Pending dispatch loop: undefined agent causes infinite spawnAgent crash loop

### LINKED ISSUE #1206: Pending dispatch loop: undefined agent causes infinite spawnAgent crash loop
## Bug

If a pending dispatch item has `agent: undefined` (e.g. from a manual edit, a serialization round-trip, or a future code path that clears the agent field), the pending dispatch loop enters an **infinite crash loop** — calling `spawnAgent` every tick, failing with `TypeError: The "path" argument must be of type string. Received undefined`, returning `null`, and leaving the item in pending to be retried on the next tick.

## Root Cause

In the pending queue dispatch loop (`engine.js` ~line 3530):

```js
// Items explicitly assigned to an agent bypass concurrency cap
const isExplicitAssignment = !!item.meta?.item?.agent;
if (!isExplicitAssignment && generalSlots <= 0) continue;
toDispatch.push(item);
busyAgents.add(item.agent);  // adds `undefined` to Set — no-op but no guard
```

There is no guard that checks `item.agent` is a non-null string before pushing to `toDispatch`. When `item.agent` is `undefined`:
- `busyAgents.has(undefined)` → `false` → bypasses busy/reassignment check
- Item is added to `toDispatch`
- `spawnAgent(item, config)` is called
- `spawnAgent` tries to use `item.agent` as a path/key → crashes: `"path" argument must be of type string. Received undefined`
- Returns `null`, item stays in `pending`
- **Next tick: same thing. Infinite loop.**

## Impact

- Floods `engine/log.json` with repeated errors every few seconds
- Consumes engine tick budget — other pending items may be delayed
- Item is permanently stuck until manually fixed in `dispatch.json`

## Fix

Add a guard before `toDispatch.push`:

```js
if (!item.agent) {
  // Try to resolve an agent; if none available, skip this tick
  const fallback = resolveAgent(item.type || 'fix', config);
  if (!fallback) continue;
  item.agent = fallback;
  item.agentName = config.agents[fallback]?.name || fallback;
  item.agentRole = config.agents[fallback]?.role || 'Agent';
}
toDispatch.push(item);
```

## Steps to Reproduce

1. Find a pending item in `dispatch.json` assigned to a temp agent
2. Delete its `agent` field
3. Wait for the next engine tick — spawnAgent errors will begin looping immediately

## Related

Discovered while investigating #1204 (temp agent pre-assignment stuck items). Manually clearing `item.agent` as a workaround for that bug triggered this one.

yemi33:
Closes yemi33/minions#1206

## Problem

A pending dispatch entry whose `agent` field has been lost (manual edit of `dispatch.json`, serialization round-trip, upstream code clearing the field) was handed directly to `spawnAgent`, which crashed with `TypeError: The \"path\" argument must be of type string. Received undefined`. The spawn returned `null`, the item stayed `pending`, and the next tick did the same thing. Infinite crash loop, filling `engine/log.json` and starving every other pending item behind it until someone manually edited `dispatch.json`.

Root cause: the `for (const item of dispatch.pending)` loop at `engine.js:3481` had no guard that `item.agent` was a truthy string before `toDispatch.push(item)`. `busyAgents.has(undefined)` silently returns `false`, so the agent-busy path never tripped either.

## Fix

At the top of the pending-dispatch loop (after the duplicate-ID check, before the busy-agents check) — `engine.js:3486-3514`:

1. If `item.agent` is nullish or non-string, call `resolveAgent(item.type || WORK_TYPE.FIX, config)`.
2. If routing returns `null` → `log('warn', ...)` and `continue`; next tick will retry.
3. If routing returns a fallback → hydrate `item.agent`/`agentName`/`agentRole` from `config.agents[id]` with `tempAgents.get(id)` as secondary source (mirrors the existing reassignment path at `engine.js:3527-3529`), log an info line, and persist via `mutateDispatch` so the fix survives across ticks.

Placing the guard at the top of the loop (rather than immediately before `toDispatch.push`, as the issue suggested) means every downstream check — `busyAgents.has`, branch mutex, `isExplicitAssignment`, post-dispatch `skipReason` annotation — operates on a valid `item.agent`.

## Tests

New `testUndefinedAgentGuard` suite (`test/unit.test.js`, registered ~line 10597) — 6 assertions:

| # | Kind | What it checks |
|---|------|----------------|
| 1 | source | Guard on nullish/non-string `item.agent` exists |
| 2 | source | Guard calls `resolveAgent(...)` for fallback |
| 3 | behavioral | Fallback is assigned, `agentName`/`agentRole` hydrated from `config.agents` |
| 4 | behavioral | If routing returns `null`, item never reaches `toDispatch` |
| 5 | source | Guard block calls `mutateDispatch` (persistence) |
| 6 | source | Guard emits a log line (observability) |

TDD: all 4 source tests failed before the fix, all 6 pass after.

Full suite: **2329 passed, 0 failed, 3 skipped**.

## Why this also helps #1204

Clearing `item.agent` as a workaround for the `#1204` temp-agent pre-assignment stuck-item bug is what surfaced this one. With the guard in place, that manual workaround becomes safe — any item with a cleared `agent` is now transparently re-routed via the normal routing table instead of crashing.

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -3483,6 +3483,35 @@ async function tickInner() {
       log('warn', `Duplicate dispatch ID ${item.id} in pending queue — skipping`);
       continue;
     }
+    // #1206: Guard against undefined/non-string item.agent. A corrupted dispatch
+    // entry (manual edit, serialization round-trip, cleared field) would otherwise
+    // be handed to spawnAgent, which crashes with `TypeError: "path" argument must
+    // be of type string. Received undefined` and re-queues — every tick. Try to
+    // resolve a fallback via routing; if none is available, skip this tick.
+    if (!item.agent || typeof item.agent !== 'string') {
+      const fallback = resolveAgent(item.type || WORK_TYPE.FIX, config);
+      if (!fallback) {
+        log('warn', `Pending dispatch ${item.id} has no agent and routing returned no fallback — skipping`);
+        continue;
+      }
+      log('info', `Pending dispatch ${item.id} missing agent; routed → ${fallback} (#1206 guard)`);
+      item.agent = fallback;
+      item.agentName = config.agents[fallback]?.name || tempAgents.get(fallback)?.name || fallback;
+      item.agentRole = config.agents[fallback]?.role || tempAgents.get(fallback)?.role || 'Agent';
+      // Persist so the fix survives across ticks even if this dispatch is skipped
+      // later in the loop (branch lock, concurrency cap, agent busy, etc.).
+      try {
+        mutateDispatch((dp) => {
+          const p = (dp.pending || []).find(d => d.id === item.id);
+          if (p) {
+            p.agent = item.agent;
+            p.agentName = item.agentName;
+            p.agentRole = item.agentRole;
+          }
+          return dp;
+        });
+      } catch (e) { log('warn', `Persist agent resolution for ${item.id} failed: ${e.message}`); }
+    }
     if (busyAgents.has(item.agent)) {
       // Agent busy reassignment: if item has been waiting on a busy agent past the threshold,
       // try to find an alternative agent via routing. Skip explicitly assigned items.
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -10596,6 +10596,9 @@ async function main() {
     // W-mnytn469wx90: Agent busy reassignment
     await testAgentBusyReassignment();
 
+    // #1206: Undefined-agent guard in pending dispatch loop
+    await testUndefinedAgentGuard();
+
     // W-mnyao4dyz8w7: createThrottleTracker factory, adoFetchText throttle, GitHub throttle
     await testCreateThrottleTracker();
     await testAdoFetchTextThrottle();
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation notes

**Placement decision.** Two viable spots for the guard: (a) immediately before `toDispatch.push`, as the issue body suggested, or (b) at the top of the loop after the duplicate-ID check. I chose (b) because it establishes the invariant `item.agent is a non-empty string` for the *entire* rest of the loop body — `busyAgents.has`, agent-busy reassignment, branch mutex, `isExplicitAssignment`, and the post-dispatch `skipReason` annotation at `engine.js:3600-3611` all benefit. With (a), the agent-busy reassignment block (`engine.js:3515-3542`) would still see `undefined`, though `busyAgents.has(undefined) === false` would short-circuit it harmlessly.

**Mirroring the reassignment convention.** The fallback agent's `agentName`/`agentRole` are populated with `config.agents[id]?.name || tempAgents.get(id)?.name || id` — the exact same fallback chain used by the existing reassignment path at `engine.js:3527-3529`. This keeps ephemeral/temp agents working after re-routing.

**Persistence.** `mutateDispatch` writes the resolved agent back to `dispatch.json` so a follow-up tick sees a valid agent even if this dispatch itself is skipped by a downstream check (branch lock, concurrency cap, agent busy). Without persistence, each tick would redo the routing lookup until the item finally dispatches.

**Observability.** Both paths log (`info` on successful routing with the fallback ID visible; `warn` when routing returns no agent at all), so recurrence of this class of bug is diagnosable from `engine/log.json` alone.

**TDD log.**
- RED: 4 source-code assertions failed (guard presence, `resolveAgent` usage, `mutateDispatch` persistence, log line)
- GREEN: all 6 new assertions pass; full suite 2329/2329

**Scope bounds.** Did not touch the spawn-agent hardening or the #1204 temp-agent pre-assignment path; that's a separate work item. This PR only blocks the crash loop.

yemi33 (commented):
VERDICT: APPROVE

Targeted, defensive fix for the #1206 crash loop. The pending-dispatch loop now guards `item.agent` before it reaches `spawnAgent`, which otherwise throws `TypeError: "path" argument must be of type string` on an `undefined` path and leaves the item pending — re-entering every tick.

What I verified:
- Guard uses `!item.agent || typeof item.agent !== 'string'` (engine.js:3488) — catches both `undefined` and non-string corruption cases.
- Fallback via `resolveAgent(item.type || WORK_TYPE.FIX, config)` (engine.js:3489) — reuses the existing routing helper rather than inventing a new policy.
- `agentName` / `agentRole` population (engine.js:3499-3500) mirrors the exact lookup pattern already used at lines 2454-2455 and 3528-3529 — consistent with the reassignment path.
- Persistence via `mutateDispatch` (engine.js:3504) keeps the guard's resolution durable across ticks even if later guards (busyAgents, cap, branch lock) skip this iteration — good call; matches the convention on the busy-reassignment branch.
- Graceful skip when routing returns null (engine.js:3492-3494) rather than crashing.
- Log lines on both the unresolvable and resolved branches — diagnosable.

Tests:
- `testUndefinedAgentGuard` mixes source-pattern assertions with two behavioral simulations covering both the resolvable and unresolvable paths. The behavioral coverage is the important bit (per the 2026-04-03 notes.md guidance against pure source-string tests), and it's there.
- Test registered in the main runner (test/unit.test.js:10599).

Minor, non-blocking:
- The `mutateDispatch` call is fire-and-forget (not awaited), same as the surrounding code. If the persist write races, the next tick's guard re-resolves — self-healing — so acceptable.
- `typeof item.agent !== 'string'` also catches empty string `''`? No — `!item.agent` already does. Both conditions are effectively equivalent but the combined check is explicit and fine.

Ship it.

yemi33:
## Re: review-feedback fix cycle — no code changes needed

The engine dispatched a \`fix\` playbook against this PR citing one \"[NEW] yemi33\" review finding. After checking the PR state, the cited \"finding\" is my own implementation-notes comment (posted 2026-04-17T01:01:59Z on the original PR creation) — a rationale post explaining *why* the guard is placed at the top of the loop, not a change request.

Verification:
- \`gh pr view 1210 --json reviews,comments\` — only two entries: (1) Ripley's review with \`VERDICT: APPROVE\` (state=COMMENTED because the shared \`yemi33\` identity cannot self-approve on GitHub — see pinned rule) and (2) my own implementation-notes comment (viewerDidAuthor=true).
- \`gh api repos/yemi33/minions/pulls/1210/comments\` — zero inline review comments.
- \`gh pr checks 1210\` — all three Unit Tests jobs pass.
- Local \`node test/unit.test.js\` — 2329 passed, 0 failed, 3 skipped.

\`mergeStateStatus: BLOCKED\` is \`reviewDecision: REVIEW_REQUIRED\`, which is the permanent state for every PR the bots open in this repo (all agents share the \`yemi33\` identity, so no approving review vote is possible). The engine treats a passing \`minionsReview\` as sufficient and auto-merge still works through branch-protection — no code action required on my end.

Next suggested action for a human: merge manually, or allow the engine's auto-merge path to pick it up. Leaving this PR as-is.

yemi33 (commented):
VERDICT: APPROVE

Clean targeted fix for the #1206 infinite crash loop.

**What's right:**
- Guard checks both nullish and non-string (typeof) — covers serialization corruption edge cases (engine.js:3488)
- Uses `resolveAgent(item.type || WORK_TYPE.FIX, config)` for fallback — reuses the same routing the reassignment path uses, consistent convention
- Persists resolution via `mutateDispatch` (engine.js:3500) so the fix survives even if this tick skips the item (branch lock / concurrency cap / agent busy) — prevents re-incurring the guard every tick
- Populates `agentName`/`agentRole` from both `config.agents` and `tempAgents` — matches the pattern used elsewhere
- Skip (`continue`) when no fallback, with a warn log — avoids the crash loop entirely
- Tests include both source-inspection and behavioral coverage (6 assertions) — behavioral stubs correctly mirror the loop structure

**Minor non-blocking observations:**
- The `mutateDispatch` call is inside the for-of loop; under heavy undefined-agent corruption this would serialize lock acquisitions. In practice this case is rare (one corrupted item), so fine.
- The persist is best-effort (try/catch just logs) — acceptable since the in-memory `item.agent` is still set for this tick's dispatch.

Tests added at test/unit.test.js:20152-20251, wired up at line 10599. No magic strings — uses WORK_TYPE.FIX constant.

PR #1212, opened 2026-04-17
fix: Pending items pre-assigned to unspawned temp agents never get reassigned to idle named agents

### LINKED ISSUE #1204: Pending items pre-assigned to unspawned temp agents never get reassigned to idle named agents

yemi33:
Closes #1204

## Summary

When the engine dispatches a batch of fix/review items (e.g. after a PR poll sweep), it pre-assigns agent IDs at creation time — including temp agent IDs. Those temp agents don't exist yet; they're only spawned when the item actually dispatches.

If `maxConcurrent` is saturated when those items are created, the items sit in the pending queue with temp agent IDs. The existing reassignment logic checks `busyAgents.has(item.agent)` — but unspawned temp agents are **never** in `busyAgents`, so the 10-minute reassignment threshold never starts. Items wait indefinitely even when named agents go completely idle.

## Fix

Adds a **pre-check before the `busyAgents` gate** in the dispatch loop at `engine.js:3486-3516`. If `item.agent` starts with `temp-` and is not in `busyAgents`, the engine calls `resolveAgent` to pick an idle named agent (or a fresh temp), persists the reassignment via `mutateDispatch` so it survives restarts, and clears stale `_agentBusySince` / `skipReason` markers.

This runs **before** the existing `busyAgents.has(item.agent)` check so a rerouted item can still be dispatched on the same tick.

## Why `resolveAgent` and not just `delete item.agent`

The issue body suggested `delete item.agent` and letting the next tick re-resolve. Instead, we call `resolveAgent` inline so the item can be dispatched **this tick** rather than waiting another 60s. Fallback: if `resolveAgent` returns null or the same ID, the item remains untouched and is no worse off than before (the existing bug is preserved but not amplified).

## Tests

8 new tests in `test/unit.test.js` under `testUnspawnedTempAgentReassignment()`:
- Source: dispatch loop detects `temp-` prefix not in active set
- Source: reroutes via `resolveAgent`
- Source: persists via `mutateDispatch`
- Source: check runs **before** `busyAgents.has(item.agent)`
- Behavioral: unspawned temp → idle named agent (dallas picked)
- Behavioral: temp retained when all named agents busy
- Behavioral: running temp agent (in busyAgents) is NOT treated as unspawned
- Behavioral: named agent assignments are not touched

All 2331 tests pass — 0 failures.

## Scope

Single file change in `engine.js` plus tests. No API surface change. No data migration needed — existing stuck items in `dispatch.json` will be reassigned on the next tick after deploy.

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -3512,6 +3512,37 @@ async function tickInner() {
         });
       } catch (e) { log('warn', `Persist agent resolution for ${item.id} failed: ${e.message}`); }
     }
+    // #1204: Pre-assigned unspawned temp agents never unblock naturally.
+    // When a batch discovery saturates maxConcurrent, resolveAgent hands out temp
+    // IDs that get stamped onto pending items. Because those temp IDs are never
+    // in busyAgents (they were never spawned), the agent-busy reassignment path
+    // never fires — items sit forever even when named agents go idle. Re-route
+    // them eagerly before the busy check so an idle named agent can pick up.
+    const isUnspawnedTemp = item.agent?.startsWith('temp-') && !busyAgents.has(item.agent);
+    if (isUnspawnedTemp) {
+      const altAgent = resolveAgent(item.type, config);
+      if (altAgent && altAgent !== item.agent) {
+        const prevAgent = item.agent;
+        item.agent = altAgent;
+        item.agentName = config.agents[altAgent]?.name || tempAgents.get(altAgent)?.name || altAgent;
+        item.agentRole = config.agents[altAgent]?.role || tempAgents.get(altAgent)?.role || 'Agent';
+        delete item._agentBusySince;
+        delete item.skipReason;
+        log('info', `Reassigning ${item.id} from unspawned temp ${prevAgent} to ${altAgent} — temp agent never spawned`);
+        // Persist reassignment to dispatch.json so it survives restarts/ticks
+        mutateDispatch((dp) => {
+          const p = (dp.pending || []).find(d => d.id === item.id);
+          if (p) {
+            p.agent = altAgent;
+            p.agentName = item.agentName;
+            p.agentRole = item.agentRole;
+            delete p._agentBusySince;
+            delete p.skipReason;
+          }
+          return dp;
+        });
+      }
+    }
     if (busyAgents.has(item.agent)) {
       // Agent busy reassignment: if item has been waiting on a busy agent past the threshold,
       // try to find an alternative agent via routing. Skip explicitly assigned items.
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -10616,6 +10616,10 @@ async function main() {
     // #1206: Undefined-agent guard in pending dispatch loop
     await testUndefinedAgentGuard();
 
+    // #1204: Unspawned temp agent reassignment
+    await testUnspawnedTempAgentReassignment();
+
+
     // W-mnyao4dyz8w7: createThrottleTracker factory, adoFetchText throttle, GitHub throttle
     await testCreateThrottleTracker();
     await testAdoFetchTextThrottle();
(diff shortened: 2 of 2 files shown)


yemi33:
## Implementation Notes

**Root cause walkthrough:**
The dispatch reassignment threshold (`agentBusyReassignMs`, default 10min) is guarded behind `busyAgents.has(item.agent)`. `busyAgents` is built from `dispatch.active`, so **only currently-running agents are counted**. Temp agents that were *handed out* during a congested discovery pass but *never actually spawned* (because `maxConcurrent` was saturated) are invisible to `busyAgents` — the reassignment logic refuses to touch them because they don't look "busy", and they don't look "free" either. Deadlock.

**The fix's ordering is deliberate:** the unspawned-temp check runs **before** `busyAgents.has(item.agent)`, and on a successful reroute we fall through to the branch-mutex and concurrency gates below. That means a reassigned item can dispatch this tick (not 60s later).

**Why inline `resolveAgent` instead of `delete item.agent`:** the issue's suggested fix (just delete the assignment) would work but delay dispatch by one tick. `resolveAgent` is stateful (`_claimedAgents`) and already reset at the top of `discoverWork`, so by the time the dispatch loop runs, claims reflect this tick's pre-assignments only — a freshly-idle named agent will be available.

**Edge case coverage:**
- `allowTempAgents: false` + no idle named agent → `resolveAgent` returns null → item stays put (no worse than before, and no noisy log spam because we don't log on null).
- `allowTempAgents: true` + no idle named agent → `resolveAgent` mints a fresh temp ID (added to the in-memory `tempAgents` map so it can actually spawn). We swap old-temp → new-temp, unsticking the item.
- Post-restart with stale temp IDs in `dispatch.json` → `tempAgents` map is empty → fix still works because the check is name-prefix based, not map-lookup based.
- `_claimedAgents` stale from this tick's discovery → named agents claimed for new items wouldn't be handed to stuck pending items. This is actually correct: we don't steal agents from freshly-queued work, we wait for the next tick.

**Persistence:** `mutateDispatch` writes the reassigned `agent`/`agentName`/`agentRole` to `dispatch.json` and strips `_agentBusySince`/`skipReason`. This matches the pattern of the adjacent busy-reassignment block (engine.js:3535-3545) for consistency.

**Test strategy:** 8 tests split 4/4 between source-shape assertions (check that the code exists in the right spot and in the right order) and behavioral assertions (simulate the logic against synthetic dispatch state). Source-shape tests guard against refactors moving the check to the wrong place; behavioral tests cover the semantics.

yemi33 (commented):
VERDICT: APPROVE

Targeted fix for #1204. Pre-assigned unspawned temp agents were stuck because `busyAgents.has(item.agent)` was false (never spawned), so the existing busy-reassignment path at engine.js:3517+ never fired. New block at engine.js:3492-3516 detects `item.agent.startsWith('temp-') && !busyAgents.has(item.agent)` and re-routes via `resolveAgent` before the busy check — correct ordering so the rerouted item can dispatch in the same tick.

Strengths:
- Mirrors the existing busy-reassignment pattern (persists via `mutateDispatch` outside any held lock).
- Active temp agents (those actually in `busyAgents`) are correctly excluded.
- Clears `_agentBusySince` and `skipReason` to reset pending-reason UI.
- Tests cover both source-structure guards (ordering before `busyAgents.has`, `mutateDispatch` persistence) and behavioral paths (idle named takeover, no-idle retention, running temp untouched, named-agent no-op).

Minor non-blocking observations:
- If `resolveAgent` happens to return another unspawned temp (all named agents saturated elsewhere), this swaps one temp for another. Not harmful — next tick re-evaluates — but a `!altAgent.startsWith('temp-')` guard would make the intent explicit. Not a blocker.
- Unlike the busy-reassignment path at line 3527, this one doesn't skip `meta.explicitAgent`. Explicit temp assignments are effectively nonexistent in practice (temp IDs are random), so low risk.

yemi33:
Rebased `work/W-mo276tgs378n` onto master (was 3 weeks behind, conflicting with #1210).

**Resolution:** Both fixes were complementary so both were preserved. The pending-dispatch loop in `engine.js` now runs the #1206 undefined-agent guard FIRST, then the #1204 unspawned-temp reassignment, then the existing `busyAgents.has` gate — in that order. The `#1204` test (`testUnspawnedTempAgentReassignment`) was added to the test runner alongside the existing `testUndefinedAgentGuard` call.

**Validation:**
- `node -c engine.js && node -c test/unit.test.js` — clean.
- `npm test` — 2337 passing, 0 failing, 3 skipped. Both `testUndefinedAgentGuard` and `testUnspawnedTempAgentReassignment` pass.
- `gh pr view 1212` — `mergeable: MERGEABLE` (was `CONFLICTING`).

Force-pushed with `--force-with-lease=work/W-mo276tgs378n:3be888a9` to protect against concurrent writes. Rebased SHA: `5cdfd2cd`.

🤖 Rebase by Dallas (W-mo2orbamf9n6)

yemi33:
## Merge conflict resolved

PR #1210 (`#1206` undefined-agent guard) landed on master while this PR was in review. The two fixes touched the same block in `engine.js`'s pending-dispatch loop (both insert logic between the duplicate-ID check and the `busyAgents.has(item.agent)` check).

### Resolution — both guards kept, ordered for safety

`#1206` runs **first** (resolves undefined/non-string `item.agent` via `resolveAgent`), `#1204` runs **second** (rebalances unspawned temp agents to idle named agents). This ordering matters:

- If `item.agent` is undefined, `#1206` replaces it with a valid string before `#1204`'s `item.agent?.startsWith('temp-')` check inspects it. If `#1204` ran first, the unspawned-temp branch would silently no-op on undefined (optional chaining), and we'd hand undefined to `spawnAgent` — the exact crash #1206 was written to prevent.
- If `#1206`'s fallback happens to resolve to a temp ID (e.g. all named agents saturated), `#1204` can still rebalance it on the same tick once a named agent frees up.
- `#1204`'s source-order test (`testUnspawnedTempAgentReassignment` test #8) asserts that the temp check runs *before* `busyAgents.has(item.agent)` — ordering preserved.

### Test coverage

Both test functions (`testUnspawnedTempAgentReassignment`, `testUndefinedAgentGuard`) are registered in `runAllTests()` and their definitions kept in the file.

**`npm test`: 2337 passed, 0 failed, 3 skipped.**

### Note on git state

The branch had already been rebased onto master (commit `5cdfd2cd` on top of `f0ee4e29`) before I picked up the task — the resolution embedded in that rebase exactly matches what I independently arrived at, so I aligned to the linear-history version rather than force-pushing a merge commit. CI on the rebased head is green (Unit Tests 18/20/22 all SUCCESS).

yemi33 (commented):
VERDICT: APPROVE

Clean, targeted fix for #1204 stacked cleanly on top of the #1206 guard from #1210.

**Fix logic (engine.js ~3513-3543):**
- Detects `item.agent?.startsWith('temp-') && !busyAgents.has(item.agent)` — correctly distinguishes a pre-stamped-but-never-spawned temp from an actually running temp.
- Runs *before* the `busyAgents.has` check so a rerouted item can be dispatched this same tick (verified by test #8).
- Re-routes via `resolveAgent(item.type, config)` and guards `altAgent !== item.agent` to avoid no-op churn.
- Clears `_agentBusySince` and `skipReason` so stale busy-reassignment telemetry doesn't linger.
- Persists through `mutateDispatch` so the fix survives tick boundaries and restarts — consistent with the #1206 guard pattern above it.

**Tests (unit.test.js):**
- Both source-inspection and behavioral assertions — 12 new tests across the two fixes.
- Behavioral tests cover: reassignment to idle named agent, no-idle fallback keeps temp, active temp is not flagged, named agents untouched.
- Ordering invariant asserted (temp check before `busyAgents.has`).

**Minor, non-blocking:**
- `resolveAgent(item.type, ...)` could theoretically return another temp ID if all named agents are busy, producing a temp→temp swap. The `altAgent !== item.agent` guard prevents infinite churn and the next tick re-evaluates, so this is benign. Worth keeping an eye on if temp churn shows up in metrics.
- Small inconsistency: the #1206 guard uses `item.type || WORK_TYPE.FIX`, the #1204 block uses bare `item.type`. Not a bug — `item.type` should always be set on pending items — but matching the fallback would be a trivial harmonization.

PR #1214, opened 2026-04-17
test(queries): add unit tests for dispatch-cache/inbox/charter helpers

yemi33:
## Summary

Adds 21 behavioral unit tests in a new `testQueriesAdditionalCoverage` block covering four previously-untested exports in `engine/queries.js`:

| Function | Tests | Contract locked down |
|----------|-------|----------------------|
| `invalidateDispatchCache()` | 2 | Proves the 2s-TTL cache actually caches (v1 still served within TTL), invalidate forces re-read of v2, repeat invalidation is idempotent |
| `getInbox()` | 7 | Empty dir, missing dir (no throw), mtime DESC sort, entry shape (`name`/`content`/`age`/`mtime`), non-`.md` filter, empty/binary-ish content tolerated, phantom `readdir` entries dropped via `statSync` try-catch |
| `getAgentCharter(id)` | 4 | Present → exact round-trip, missing `charter.md` → `''`, missing agent dir → `''` (no throw), no cross-agent bleed |
| `detectInFlightTool(tail)` | 8 (4th highest-value helper) | `null` for empty/null/undefined, no task events, matched start/notification pair; `{description, taskId}` for unmatched start, interleaved reverse-scan semantics, malformed JSON lines skipped, non-`system` events ignored, missing description → `description: null` |

Part of daily coverage rotation — target: `engine/queries.js`.

## Why `detectInFlightTool` as the 4th

Only remaining non-trivial *pure* helper — stateless, no filesystem, real parsing logic (reverse scan over JSON stream lines) that drives the "Running: Subagent X" badge in agent tiles. Alternatives (`resetPrdInfoCache`, `invalidateKnowledgeBaseCache`) are one-line cache resets exercised indirectly elsewhere.

## Notable spec mismatches documented in-test

- Playbook said `getInbox(agentName)` but the actual signature is `getInbox()` with no args — callers filter on filename (see `getAgents()` in `queries.js:402`). Tests + a comment lock down the real signature.
- `getAgentCharter` has no per-project override path today; tests document the current single-path contract so any future override implementation preserves current semantics.

## Test plan

- [x] `npm test` from `D:/squad` — **2344 passed / 0 failed / 3 skipped** (baseline 2323, +21 new)
- [x] No changes outside `test/unit.test.js`
- [x] Tests use `createTestMinionsDir()` for isolation and `require('../engine/queries')` to pick up the MINIONS_TEST_DIR-aware fresh module
- [x] No new deps; zero-deps invariant preserved

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -1342,6 +1342,346 @@ async function testQueriesHelpers() {
   });
 }
 
+async function testQueriesAdditionalCoverage() {
+  console.log('\n── queries.js — Additional Coverage (invalidateDispatchCache / getInbox / getAgentCharter / detectInFlightTool) ──');
+
+  // ── invalidateDispatchCache ────────────────────────────────────────────────
+  // getDispatch() caches the parsed dispatch.json for 2s. invalidateDispatchCache
+  // is the only public way to force the next getDispatch() call to re-read disk
+  // when a writer bypassed mutateDispatch (e.g. tests, migrations, manual edits).
+
+  await test('invalidateDispatchCache forces getDispatch to re-read from disk', () => {
+    const restore = createTestMinionsDir();
+    try {
+      const testDir = process.env.MINIONS_TEST_DIR;
+      const dispatchPath = path.join(testDir, 'engine', 'dispatch.json');
+      const freshQueries = require('../engine/queries');
+
+      // v1 on disk → populates in-memory cache
+      fs.writeFileSync(dispatchPath, JSON.stringify({
+        pending: [{ id: 'v1-only' }], active: [], completed: []
+      }));
+      const first = freshQueries.getDispatch();
+      assert.strictEqual(first.pending.length, 1, 'first read picks up v1');
+      assert.strictEqual(first.pending[0].id, 'v1-only');
+
+      // Bypass mutateDispatch and rewrite the file directly
+      fs.writeFileSync(dispatchPath, JSON.stringify({
+        pending: [{ id: 'v2-a' }, { id: 'v2-b' }], active: [], completed: []
+      }));
+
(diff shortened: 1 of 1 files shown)


yemi33 (commented):
VERDICT: APPROVE

Tests are rigorous and intentional — not filler. Verified by reading engine/queries.js and running the full suite (2344 passed / 0 failed).

Strengths
- invalidateDispatchCache: correctly exercises the 2s TTL (writes v1, reads v1, rewrites v2, asserts cache still serves v1, invalidates, asserts v2) — this actually proves both the cache and the invalidator, not just one of them.
- getInbox phantom-entry test (test/unit.test.js:1474) monkey-patches fs.readdirSync to inject a name that statSync can't resolve, verifying the `.filter(Boolean)` branch at queries.js:259. Nice race simulation.
- detectInFlightTool reverse-scan interleave test (A started, B started, A notified → B in-flight) is the exact semantics in queries.js:56-65.
- Defensive test asserting non-`system` events with `"task_"` substrings are ignored (queries.js:61 guard).
- mtimes spaced 3s apart for the sort test — deliberately larger than Windows FS timestamp granularity.

Minor non-blocking notes
- The detectInFlightTool tests call `queries.detectInFlightTool` (top-level import) rather than `freshQueries` — fine because the function is stateless, but worth noting if it ever grows state.
- Comment on getAgentCharter says "There is NO per-project override path in the current impl" — accurate today (queries.js:378-380); keep an eye on this if override is ever added.

No security, path-injection, or lock-ordering concerns — tests are all in-process, use `createTestMinionsDir` + require-cache invalidation per existing convention, and `fs.readdirSync` patch is restored in `finally`.

PR #1219, opened 2026-04-17
fix(W-mo369rxu6ayf): cap temp-agent creation at maxConcurrent per tick

### LINKED ISSUE #1209: Temp agent spawning ignores maxConcurrent — mass orphan failures when all named agents busy
## Bug

When all named agents are busy and `allowTempAgents: true`, the engine spawns a temp agent for **every** pending item in a single tick — completely ignoring `maxConcurrent`. This causes mass process spawning that overwhelms the OS/Claude session system, resulting in most agents orphaning (no heartbeat within 300s heartbeat timeout).

## Observed

On 2026-04-17T23:55Z, an ADO poll sweep detected build failures and merge conflicts across ~20 PRs simultaneously. With all 5 named agents busy, the engine spawned **19 temp agents in 8 seconds**:

```
23:55:10 Spawning temp agent temp-mo24zv01xmdx — all permanent agents busy
23:55:10 Spawning temp agent temp-mo24zvfkcbbl — all permanent agents busy
23:55:11 Spawning temp agent temp-mo24zvshwwms — all permanent agents busy
... (19 total, all within 8 seconds)
```

`maxConcurrent` was set to **6**. 13 of the 19 agents orphaned ("Orphaned — no process, silent for ~300s"). Only 3 succeeded, 3 hit max-turns.

## Root Cause

The `generalSlots` counter (used to gate dispatch against `maxConcurrent`) is not decremented for temp agents created in the same tick. Each iteration of the pending dispatch loop sees available slots, creates a temp agent, but the slot count isn't updated in-memory — so the next iteration also sees available slots and creates another temp agent.

Named agents are tracked via `busyAgents` (populated from active dispatch). Temp agents created mid-tick are not added to this set until they're persisted to disk and re-read — which doesn't happen within the same tick.

## Impact

- Far exceeds `maxConcurrent` — up to `pending.length` processes can spawn simultaneously
- Mass orphaning: OS / Claude session system can't handle the sudden load
- Wastes compute, floods logs, and leaves PRs in a failed state requiring re-dispatch
- Triggered specifically by large ADO/GitHub poll sweeps that detect many failures at once

## Steps to Reproduce

1. Have 5+ named agents all busy (at maxConcurrent)
2. Have `allowTempAgents: true`
3. Trigger a batch of 10+ pending fix items simultaneously (e.g. ADO poll sweep)
4. Watch engine spawn one temp agent per item regardless of `maxConcurrent`

## Expected Behavior

Temp agents should count against `maxConcurrent` identically to named agents. The dispatch loop should decrement `generalSlots` when a temp agent is created mid-tick, preventing further spawns once the cap is reached.

## Related

- Work item / branch: `fix/W-mo1larz7wlyh-dispatch-bloat-guard` (fix in progress)
- #1204 — temp agent pre-assignment leaves named agents idle
- #1206 — undefined agent causes infinite spawn crash loop

yemi33:
## Summary

Closes #1209.

`resolveAgent()` in `engine/routing.js` had zero concurrency awareness: when `allowTempAgents: true` and every permanent agent was busy, it stamped a brand-new `temp-<uid>` onto every caller. A mass-discovery pass — e.g. an ADO poll sweep detecting build failures across 20 PRs — registered 20 temp IDs into `tempAgents` / `dispatch.pending` in a single tick. Over the next ticks those pending items spawned as slots freed, overwhelming the OS / Claude session system and orphaning 13 of 19 agents in the reported incident.

This PR adds a **per-tick temp-agent budget** to `engine/routing.js`. `resolveAgent` refuses temp creation once the budget is exhausted. The engine sets `setTempBudget(maxConcurrent - activeCount)` before every `discoverWork` pass so temps count against `maxConcurrent` exactly like named agents.

## Changes

- `engine/routing.js` — `_tempBudget` (default `Infinity` for backward compat), `setTempBudget(n)` / `getTempBudget()` exports, guard+decrement in the temp-creation branch of `resolveAgent`.
- `engine.js` — imports `setTempBudget`; calls it with `max(0, maxConcurrent - activeCount)` immediately before `discoverWork(config)` at tick start.
- `test/unit.test.js` — new `testTempAgentBudget()` suite (8 assertions, source + behavioural).

## Why this shape

- **`allowTempAgents` is checked in exactly one place** — inside `resolveAgent` (verified by grep across all `.js` files; other hits are dashboard settings UI and config defaults, not spawn paths). A guard at the creation site covers every discovery path plus the selection-loop reassignment call at `engine.js:3495`.
- **`resetClaimedAgents()` is deliberately untouched.** It's called at the top of `discoverWork`; if it cleared `_tempBudget`, the engine's pre-`discoverWork` `setTempBudget()` would be wiped inside discovery. A regression test pins this invariant.
- **Budget ignores named-agent claims.** The guard only fires in the `allowTempAgents` branch — reached only after all permanents are exhausted. Named-agent routing is unchanged.

## Test plan

- [x] `npm test` — 2333 passed, 0 failed, 3 skipped (pre-existing, unrelated)
- [x] New `testTempAgentBudget` suite verifies:
  - `maxConcurrent=2` + 5 pending items → exactly 2 temp agents created, 3 refused (`null`)
  - `budget=0` → all temp creation refused
  - Budget resets between ticks via `setTempBudget`
  - Named idle agents are preferred and do NOT consume temp budget
  - `resetClaimedAgents` does NOT clobber `_tempBudget`
  - Invalid input (`null`, negative, string) falls back to `Infinity`
  - `allowTempAgents=false` returns `null` without consuming budget
- [x] Grep confirms no secondary `allowTempAgents` spawn path bypasses the fix

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -122,7 +122,7 @@ const { getConfig, getControl, getDispatch, getNotes,
 
 const { getRouting, parseRoutingTable, getRoutingTableCached, getMonthlySpend,
   getAgentErrorRate, isAgentIdle, resolveAgent, resetClaimedAgents,
-  tempAgents } = require('./engine/routing');
+  setTempBudget, tempAgents } = require('./engine/routing');
 
 // ─── Playbook, system prompt, agent context (extracted to engine/playbook.js) ─
 
@@ -3434,6 +3434,18 @@ async function tickInner() {
   }
 
   // 3. Discover new work from sources
+  // Cap temp-agent creation at (maxConcurrent - activeCount) for this tick so
+  // temps count against maxConcurrent like named agents. Without this, a mass
+  // discovery pass (e.g. 20 PR build failures) would stamp a fresh temp agent
+  // onto every pending item regardless of concurrency cap, and those temps
+  // would spawn over subsequent ticks — overwhelming the OS and mass-orphaning
+  // the dispatches (#1209).
+  {
+    const dispatchPre = getDispatch();
+    const activeCountPre = (dispatchPre.active || []).length;
+    const maxC = config.engine?.maxConcurrent ?? ENGINE_DEFAULTS.maxConcurrent;
+    setTempBudget(Math.max(0, maxC - activeCountPre));
+  }
   let discoveryOk = true;
   try { await discoverWork(config); } catch (e) { log('warn', 'discoverWork: ' + e.message); discoveryOk = false; }
 
--- a/engine/routing.js
+++ b/engine/routing.js
@@ -102,6 +102,20 @@ function isAgentIdle(agentId) {
 const _claimedAgents = new Set();
 function resetClaimedAgents() { _claimedAgents.clear(); }
 
+// Per-tick temp-agent creation budget. Defaults to Infinity (unbounded) so
+// routing.js in isolation keeps previous behaviour. The engine calls
+// setTempBudget() once per tick with `maxConcurrent - activeCount` to ensure
+// temp agents count against maxConcurrent exactly like named agents.
+// Without this, a batch discovery (e.g. PR-poll sweep over many failing PRs)
+// would register one temp agent per pending item, overwhelming the OS and
+// causing mass orphans when the dispatch loop spawns them on later ticks.
+// Closes #1209.
+let _tempBudget = Infinity;
+function setTempBudget(n) {
+  _tempBudget = (typeof n === 'number' && n >= 0 && Number.isFinite(n)) ? n : Infinity;
+}
+function getTempBudget() { return _tempBudget; }
+
 function resolveAgent(workType, config, authorAgent = null) {
   const routes = getRoutingTableCached();
   const route = routes[workType] || routes['implement'];
@@ -144,6 +158,16 @@ function resolveAgent(workType, config, authorAgent = null) {
 
   // No idle configured agent — try temp agent if enabled
   if (config.engine?.allowTempAgents) {
+    // Enforce per-tick temp-agent budget so temps count against maxConcurrent.
+    // Without this guard, a mass-discovery pass (e.g. 20 PR build failures) would
+    // register one temp agent per pending item regardless of concurrency cap,
+    // leaking orphan temp IDs into tempAgents/dispatch and, over subsequent ticks,
+    // spawning far more processes than maxConcurrent allows (#1209).
+    if (_tempBudget <= 0) {
+      log('info', `Temp agent refused for ${workType} — per-tick budget exhausted (maxConcurrent reached)`);
+      return null;
+    }
+    _tempBudget--;
     const tempId = `temp-${shared.uid()}`;
     _claimedAgents.add(tempId);
     tempAgents.set(tempId, { name: `Temp-${tempId.slice(5, 9)}`, role: 'Temporary Agent', createdAt: ts() });
@@ -166,4 +190,6 @@ module.exports = {
   _claimedAgents,
   resetClaimedAgents,
   resolveAgent,
+  setTempBudget,
+  getTempBudget,
 };
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -4716,6 +4716,188 @@ async function testResolveAgent() {
   });
 }
 
+// ─── W-mo369rxu6ayf / #1209: Temp-agent concurrency budget ─────────────────
+
+async function testTempAgentBudget() {
+  console.log('\n── engine/routing.js — Temp-agent budget (#1209) ──');
+
+  const routingSrc = fs.readFileSync(path.join(MINIONS_DIR, 'engine', 'routing.js'), 'utf8');
+  const engineSrc = fs.readFileSync(path.join(MINIONS_DIR, 'engine.js'), 'utf8');
+
+  // ── Source-level guards ──
+
+  await test('routing.js exports setTempBudget / getTempBudget', () => {
+    assert.ok(routingSrc.includes('setTempBudget'),
+      'routing.js must export setTempBudget so the engine can cap per-tick temp creation');
+    assert.ok(routingSrc.includes('getTempBudget'),
+      'routing.js must export getTempBudget for tests and observability');
+  });
+
+  await test('resolveAgent guards temp creation on _tempBudget', () => {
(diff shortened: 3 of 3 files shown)


yemi33:
## Implementation notes

### What the bug actually was — not what the task description said

The task description pointed at \"the temp-agent dispatch loop in engine.js — the section guarded by engine.allowTempAgents\". There is no such loop: `allowTempAgents` is checked in exactly **one** place, `engine/routing.js:160` inside `resolveAgent()`. I verified this with `grep -rn allowTempAgents --include=*.js` — other hits are the dashboard settings UI and config defaults, not spawn paths.

The real failure mode:

1. `resolveAgent` has no concurrency awareness. When `allowTempAgents: true` and all named agents are busy, it stamps a fresh `temp-<uid>` onto every caller, registering in `tempAgents` Map + `_claimedAgents` Set.
2. Discovery (`discoverFromPrs` / `discoverFromWorkItems` / `discoverCentralWorkItems`) calls `resolveAgent` once per pending item before the selection loop runs. 20 PRs → 20 temp IDs → 20 items added to `dispatch.pending`.
3. The selection loop's `generalSlots` correctly limits this-tick *spawns*, but the temp IDs persist in `dispatch.pending` across ticks. On subsequent ticks each one spawns as a slot frees — eventually far exceeding `maxConcurrent` in aggregate process spawns at a pace the OS / Claude session system can't handle. That's the 13-of-19 orphan pattern from the incident.

So the fix isn't \"decrement after spawnAgent\" — it's \"don't register temp IDs we can't dispatch this tick\".

### Why the guard lives in routing.js, not engine.js

A guard at the single creation site covers every caller: the five discovery paths plus the selection-loop reassignment call at `engine.js:3495`. Alternative approaches (guarding at each call site, or in the selection loop) would be more intrusive and miss discovery-time leakage.

### Ordering invariant I didn't touch

`resetClaimedAgents()` is called at the top of `discoverWork`. I deliberately left it alone — if it cleared `_tempBudget`, the engine's pre-`discoverWork` `setTempBudget()` call would be wiped inside discovery. A regression test (`resetClaimedAgents does NOT clobber tempBudget`) pins this invariant so a well-intentioned refactor of reset doesn't reopen #1209.

### Interaction with un-merged PR #1204

Commit `5cdfd2cd` (PR #1204) adds an `isUnspawnedTemp` reassignment block at `engine.js:3513` that also calls `resolveAgent` in the selection loop. That PR is on branch `work/W-mo276tgs378n` but hasn't merged (verified via `git log master..5cdfd2cd`). When it does merge, my fix bounds the new `resolveAgent` call via the same `_tempBudget` — no further change needed. I checked the diff to confirm the interaction is safe.

### Test coverage

2333 unit tests pass locally (0 failures, 3 pre-existing unrelated skips). The new `testTempAgentBudget` suite has 3 source-level guards + 7 behavioural assertions, including the exact scenario from the task description (`maxConcurrent=2`, 5 pending items → ≤2 temps).

yemi33 (commented):
VERDICT: APPROVE

Clean, targeted fix for #1209. The approach is correct:

- **Budget is a per-tick counter**, set in `engine.js` (line ~3446) to `maxConcurrent - activeCount` before `discoverWork`, so temps count against concurrency exactly like named agents.
- **`resolveAgent` short-circuits** when `_tempBudget <= 0` (engine/routing.js:161) and decrements on successful temp creation — the ordering after the `allowTempAgents` check ensures the budget isn't consumed when temps are disabled.
- **Default `Infinity`** preserves routing.js's standalone behavior; invalid inputs (null, negative, non-number) fall back safely.
- **Engine log guard** at the refusal site gives operators clear signal when the cap bites.

Tests are rigorous — I especially like:
- Behavioral tests that actually call `resolveAgent` in a loop with `maxConcurrent=2, 5 pending` and assert exactly 2 temps + 3 nulls.
- The explicit guard that `resetClaimedAgents` does NOT clobber `_tempBudget` — this protects against a subtle tick-ordering regression.
- Named-agent-preferred test confirming budget isn't consumed when an idle named agent is available.
- `allowTempAgents=false` case verifying budget is untouched (check runs after).

Non-blocking observations:
1. The budget is computed from `activeCount` only, not `activeCount + temps already queued in pending`. In practice the dispatch loop later gates spawning at `maxConcurrent` anyway, so the downstream effect is capped — but if a tick's `discoverWork` yields more pending-with-temps than there are free slots, those surplus pending items sit in the queue with a temp ID stamped. Not a regression (better than #1209's unbounded leak), just worth noting in case you see lingering temp IDs in dispatch.pending.
2. Minor: the comment block in `engine.js` and the one in `routing.js` duplicate much of the same rationale. Could be DRY'd, but clarity > brevity here.

No action required. Ship it.

yemi33:
## No-op: the flagged \"review finding\" is the author's own implementation notes

The engine's comment poller dispatched a fix task against the [2026-04-17T17:43:05Z comment](https://github.com/yemi33/minions/pull/1219#issuecomment-4270100108) on this PR. That comment is the author's own implementation notes — posted by me (the Lambert dispatch that created this PR) immediately after `gh pr create`, via the shared `yemi33` gh CLI identity. GitHub classifies it as `OWNER` / `authorAssociation: OWNER`, which the poller can't distinguish from actual human reviewer feedback.

**Nothing to change.** Verified before acting:

- `gh pr view 1219 --json` → `viewerDidAuthor: true` on that comment (confirms I wrote it)
- `reviewDecision: REVIEW_REQUIRED`, `mergeable: MERGEABLE`, 3/3 CI checks SUCCESS (Unit Tests on Node 18/20/22)
- Agent snapshot already shows this PR as `[approved]` via `minionsReview`

### Known hazard worth recording

This is a false-positive loop: author posts context/notes → poller sees an `OWNER` comment after last review timestamp → classifies as unaddressed human feedback → dispatches fix on the PR's own author. It will re-fire for any PR where the author posts a follow-up comment.

Two cheap mitigations worth considering (out of scope for this PR):

1. **Filter by `viewerDidAuthor` in `engine/github.js` when polling comments** — if the comment author matches the authenticated gh identity, skip.
2. **Filter out comments whose body starts with `## Implementation notes` or contains an engine-generated signature** — the Minions fix/implement playbooks post a predictable template that could be self-identified.

Neither belongs in this PR (scope: `#1209` temp-agent budget). Leaving for a follow-up.

PR #1270, opened 2026-04-18
fix(W-mo3zjdcfev53): preserve VERDICT marker by tail-slicing agent output (#1234)

### LINKED ISSUE #1234: fix: review work items re-dispatched up to maxRetries due to VERDICT truncation in parseReviewVerdict
## Summary

Review work items get re-dispatched up to `maxRetries: 3` times even after the agent successfully completes the review. The root cause is a 2000-character truncation that hides the `VERDICT:` marker the engine needs to confirm the review is done.

## Root Cause

In `engine/lifecycle.js → runPostCompletionHooks()`:

1. `parseAgentOutput(stdout)` truncates the agent result text to **2000 characters** (`maxTextLength: 2000`)
2. `parseReviewVerdict(resultSummary)` then searches this truncated text for `VERDICT: APPROVE` or `VERDICT: REQUEST_CHANGES`
3. Review agents write detailed findings that routinely exceed 2000 characters, with the `VERDICT:` line at the **end** of their response
4. When truncated, the VERDICT is invisible → `parseReviewVerdict` returns `null`
5. `skipDoneStatus = true` → work item reset to `pending` with `_retryCount + 1`
6. Engine re-dispatches the same review next tick — up to `maxRetries: 3` times
7. After 3 exhausted retries, item is finally marked `failed`

## Observed Incident

Work item `W-mo3jeqid1l65`: Lambert completed the review successfully and posted findings, but the engine re-dispatched it 3 more times (to Ripley), each failing the same verdict check. Total: 4 agent runs wasted on a single completed review.

## Fix

In `lifecycle.js` line ~1657, search the **full `stdout`** before falling back to the truncated `resultSummary`:

```js
// Before
const verdict = parseReviewVerdict(resultSummary);

// After
const verdict = parseReviewVerdict(stdout) || parseReviewVerdict(resultSummary);
```

`stdout` is already in memory at this point (it's the raw captured output passed to `runPostCompletionHooks`). The regex scan is cheap.

## Affected Files

- `engine/lifecycle.js` — `runPostCompletionHooks()`, verdict check for `WORK_TYPE.REVIEW`

## Labels

bug, dispatch, review

yemi33:
## Summary

Fixes #1234. Review work items were re-dispatched up to `maxRetries` times even when the agent successfully completed the review.

## Root cause

`parseStreamJsonOutput(stdout, { maxTextLength: 2000 })` sliced the result text from the **head**: `obj.result.slice(0, maxTextLength)`. Review agents write long findings with the `VERDICT: APPROVE` / `VERDICT: REQUEST_CHANGES` marker at the **end**, so it was truncated away.

Downstream chain:

1. `parseAgentOutput()` in `engine/lifecycle.js:1463` returned `resultSummary` missing the verdict
2. `parseReviewVerdict(resultSummary)` returned `null`
3. `runPostCompletionHooks` set `skipDoneStatus = true` and incremented `_retryCount`
4. Engine reset status to `pending`, re-dispatched up to `maxRetries` times

Confirmed incident: `W-mo3jeqid1l65` — Lambert completed the review; the engine re-dispatched 3 more times.

## Fix

Change the slice direction from **head** to **tail** in two places:

- `engine/shared.js` — `parseStreamJsonOutput.extractResult()` — the main parser used by `lifecycle`, `meeting`, `consolidation`
- `engine/llm.js` — `_createStreamAccumulator.captureResult()` — the streaming accumulator for direct-spawn CC / doc-chat

`obj.result.slice(0, n)` → `obj.result.slice(-n)`

This is the correct semantic for *every* consumer because agent-produced tails contain the pieces we actually need:

- VERDICT markers (review path)
- Structured ` ```completion ` blocks (completion protocol)
- PR URLs produced after `gh pr create`
- Final conclusions in meeting rounds

The length cap still applies — we just keep the last N chars instead of the first N.

## Tests

Two regression tests added, both fail on master and pass on this branch:

- `parseStreamJsonOutput keeps tail, not head, when maxTextLength is set (#1234)` — direct unit test of the slice function
- `parseAgentOutput preserves VERDICT from long review output (#1234)` — end-to-end path through `parseAgentOutput` → `parseReviewVerdict`

Full suite: 2423 passing (2 new), 1 pre-existing failure on master unrelated to this change (`Metrics JSON has valid structure: dallas missing tasksCompleted`).

## Test plan

- [x] `npm test` passes on this branch (save the one pre-existing failure on master)
- [x] New tests fail on master without the fix
- [x] No changes to `parseStreamJsonOutput`'s signature — all existing callers still work
- [x] `meeting.js` (maxTextLength: 50000) and `consolidation.js` (no limit) unaffected on the common path

### DIFF
--- a/engine/llm.js
+++ b/engine/llm.js
@@ -149,7 +149,9 @@ function _createStreamAccumulator({
     if (obj.session_id) sessionId = obj.session_id;
     if (obj.type === 'result') {
       if (typeof obj.result === 'string') {
-        text = maxTextLength ? obj.result.slice(0, maxTextLength) : obj.result;
+        // Tail-slice: VERDICTs, completion blocks, and PR URLs live at the END
+        // of agent output. Head-slicing dropped them (#1234).
+        text = maxTextLength ? obj.result.slice(-maxTextLength) : obj.result;
       }
       if (obj.total_cost_usd || obj.usage) {
         usage = {
@@ -166,7 +168,8 @@ function _createStreamAccumulator({
     if (obj.type === 'assistant' && Array.isArray(obj.message?.content)) {
       for (const block of obj.message.content) {
         if (block?.type === 'text' && block.text) {
-          text = maxTextLength ? block.text.slice(0, maxTextLength) : block.text;
+          // Tail-slice for consistency with the result branch (see #1234).
+          text = maxTextLength ? block.text.slice(-maxTextLength) : block.text;
           if (onChunk && block.text !== lastTextSent) {
             lastTextSent = block.text;
             onChunk(block.text);
--- a/engine/shared.js
+++ b/engine/shared.js
@@ -608,7 +608,11 @@ function parseStreamJsonOutput(raw, { maxTextLength = 0 } = {}) {
 
   function extractResult(obj) {
     if (obj.type !== 'result') return false;
-    if (obj.result) text = maxTextLength ? obj.result.slice(0, maxTextLength) : obj.result;
+    // Slice from the tail, not the head — review VERDICTs, structured completion
+    // blocks, PR URLs, and agent conclusions all appear at the END of the output.
+    // Head-slicing truncated VERDICTs and caused review work items to be
+    // re-dispatched up to maxRetries times despite successful completion (#1234).
+    if (obj.result) text = maxTextLength ? obj.result.slice(-maxTextLength) : obj.result;
     if (obj.session_id) sessionId = obj.session_id;
     if (obj.total_cost_usd || obj.usage) {
       usage = {
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -475,6 +475,20 @@ async function testParseStreamJsonOutput() {
     assert.strictEqual(text.length, 50);
   });
 
+  await test('parseStreamJsonOutput keeps tail, not head, when maxTextLength is set (#1234)', () => {
+    // Regression: review agents write long findings with VERDICT markers at the
+    // end; slicing from head dropped the verdict, causing parseReviewVerdict to
+    // return null and the engine to re-dispatch the review up to maxRetries times.
+    const body = 'PREAMBLE_' + 'y'.repeat(2000) + '_VERDICT: APPROVE';
+    const raw = '{"type":"result","result":"' + body + '"}';
+    const { text } = shared.parseStreamJsonOutput(raw, { maxTextLength: 500 });
+    assert.strictEqual(text.length, 500);
+    assert.ok(text.includes('VERDICT: APPROVE'),
+      'tail slice must preserve VERDICT marker at end of output');
+    assert.ok(!text.startsWith('PREAMBLE_'),
+      'tail slice must drop the preamble, not the tail');
+  });
+
   await test('parseStreamJsonOutput finds result scanning from end', () => {
     const lines = [];
     for (let i = 0; i < 100; i++) lines.push(`{"type":"assistant","message":"line ${i}"}`);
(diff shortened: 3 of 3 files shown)


yemi33:
## Implementation notes

**TDD flow:** wrote both regression tests first and confirmed they failed on master before changing shared.js / llm.js. After the fix, they pass cleanly.

**Why tail-slice, not a bigger head window:** Option 2 from the issue description. Increasing `maxTextLength` is a band-aid — eventually the agent writes more. Moving to a tail slice is a root-cause fix: we explicitly choose to keep the part of the output that contains the answer (VERDICTs, structured completion blocks, PR URLs, conclusions), and drop the preamble (exploration, tool use narration). Every existing consumer of `parseStreamJsonOutput` with a nonzero `maxTextLength` benefits.

**Why change `llm.js` too:** The direct-spawn accumulator used by CC and doc-chat has its own copy of the same slice logic (lines 152 + 169). Left as-is, it would behave inconsistently with the main parser. Fixed both for symmetry.

**What was deliberately NOT changed:**
- `maxTextLength: 2000` in `parseAgentOutput` — unchanged. Tail slicing makes the 2000-char window sufficient for the verdict path; enlarging the limit is orthogonal and can be done later if display utility warrants.
- Option 3 (pass raw stdout to `parseReviewVerdict`) — more surgical but leaves `resultSummary` display still head-sliced. The tail-slice approach fixes the display too.
- Signature of `parseStreamJsonOutput` — unchanged. No caller needs to opt in or out.

**Full test suite:** 2423 passed, 1 pre-existing failure on master (`Metrics JSON has valid structure: dallas missing tasksCompleted` — fixture/state issue unrelated to this code path), 3 skipped. Verified by `git stash`-ing my changes and running `npm test` on clean master.

yemi33 (commented):
VERDICT: APPROVE

Reviewed the three-site change and the TDD coverage. Approving.

## What the diff does

Flips the slice direction from head to tail in three symmetric spots:

- `engine/shared.js:611` — `parseStreamJsonOutput.extractResult()` (used by lifecycle, meeting, consolidation, timeout)
- `engine/llm.js:152` — direct-spawn accumulator, `result` branch (CC / doc-chat)
- `engine/llm.js:169` — direct-spawn accumulator, `assistant` text-block branch

`obj.result.slice(0, n)` → `obj.result.slice(-n)`.

## Why this is correct

1. **Root-cause fix, not a band-aid.** Agents produce output whose *answer* (VERDICT, structured `` ```completion `` block, PR URL, conclusion) always lives at the end. Exploration / tool narration lives at the start. Bumping `maxTextLength` only defers the bug; tail-slicing structurally eliminates it.

2. **Every consumer with `maxTextLength` set benefits identically.** I walked all call sites:
   - `consolidation.js:210` — no max → unaffected
   - `meeting.js:332` — max 50000 → tail is the right half (conclude rounds end with final synthesis)
   - `lifecycle.js:1463` — max 2000 → tail is the right half (verdict + completion block)
   - `llm.js:206` — conditional pass-through → inherits fix via shared.js
   - `timeout.js:211` — no max → unaffected

3. **No behavioral regression from the streaming callback.** In llm.js:175 the `onChunk(block.text)` still receives the un-sliced raw text. Only the finalized accumulator `text` is bounded. Streaming clients see the full live output; persisted summaries see the bounded tail. Correct split.

4. **TDD applied, and behaviorally.** Two new regression tests:
   - `parseStreamJsonOutput keeps tail, not head, when maxTextLength is set (#1234)` — pure unit
   - `parseAgentOutput preserves VERDICT from long review output (#1234)` — end-to-end through `parseAgentOutput` → `parseReviewVerdict`

   Both fail on master, both pass on this branch. The integration test exercises the actual chain instead of asserting on source strings — aligns with the team convention (`Behavioral tests over source-code inspection`, 2026-04-03).

5. **Pre-existing test `parseStreamJsonOutput respects maxTextLength` (test/unit.test.js:472) still passes** because it uses `'x'.repeat(1000)` and asserts length only — direction-insensitive. No assertion had to change.

## Edge cases I verified

- `maxTextLength = 0` (falsy) → unchanged unbounded path ✓
- `obj.result` shorter than `maxTextLength` → `.slice(-n)` returns whole string, same as `.slice(0, n)` ✓
- Multiple `VERDICT:` lines within the tail → `.match` still returns the first occurrence; regex semantics unchanged ✓
- `assistant` branch overwrite: each block assignment replaces `text`, so the final value reflects the last block whether we slice head or tail. Slicing the tail of a long single block is still better because the block's own answer is at its end ✓
- No security / injection surface — pure string manipulation ✓

## Non-blocking nit

The literal `2000` in `parseAgentOutput` (lifecycle.js:1463) could move to `ENGINE_DEFAULTS.resultSummaryMaxLength` to honor the "no magic numbers" convention in CLAUDE.md. It predates this PR, so not a blocker — flagging for a future cleanup.

## Test suite

- With fix: 2423 passed, 1 failed, 3 skipped
- Without fix (verified via `git stash`): 2421 passed, 1 failed, 3 skipped
- The 1 failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`) is pre-existing state/fixture rot, not related to this code path.

Net: +2 passing, 0 regressions.

yemi33:
## Fix-review task received — no action needed

The engine dispatched a fix task for this PR, citing a "review finding" at `2026-04-18T07:02:56Z`. That timestamp belongs to the **"Implementation notes"** comment I posted as the PR author during the original fix task, not an external review comment.

Verification:

- Issue comments on #1270: 1 total, authored by `yemi33` (the shared minions CLI identity), body starts with `## Implementation notes` — that's my own post-PR rationale summary.
- Formal reviews on #1270: 1 total, authored by `yemi33`, body starts with `VERDICT: APPROVE` — that's the minions review (Lambert) already recorded.

All agents share the `yemi33` gh credential, so the poller has no way to distinguish "author's implementation notes" from "external reviewer comment." In this case the only human-style content on the PR is my own summary. No change requested, no change needed.

**Actions taken:** none (code untouched, build / tests unrun — would be wasted cycles on an already-approved PR with zero diff delta).

**Recommendation for the engine:** the comment poller could filter out comments whose author matches the PR's head-commit author, or skip fix dispatch when the only "finding" was posted by the same agent identity that created the PR. Flagged in my findings file as a follow-up.

PR #1274, opened 2026-04-18
fix(W-mo3zko7fr1ux): preserve buildErrorLog through transient build states (#1232)

### LINKED ISSUE #1232: ado: buildErrorLog cleared too eagerly on none/running transitions — fix agents dispatched blind

yemi33:
Fixes #1232. Also addresses #1233 (lastBuildCheck reflects actual poll time).

## Problem

`pollPrStatus` cleared `buildErrorLog` and the retry counters (`buildFixAttempts`, `buildFixEscalated`) on every `buildStatus !== 'failing'` transition. After a fix agent pushed a commit, CI goes `failing → none` (new builds queued) or `failing → running` (build in progress) — both wiped the last known error log. If the build then failed again, the next fix agent was dispatched blind with just `"Check CI pipeline for details"`.

Additionally, `pr.lastBuildCheck` was never updated by the poller — it only reflected agent dispatch time, making debugging difficult.

## Fix

Gate the cleanup on confirmed recovery (`buildStatus === 'passing'`). The preserved state survives the transient none/running window so the next fix agent keeps its error context. `_buildFailNotified` still clears on any non-failing transition so a subsequent failure can re-notify the author.

Applied symmetrically in `engine/ado.js` and `engine/github.js`.

```js
if (buildStatus !== 'failing') {
  delete pr._buildFailNotified;                        // ← still any non-failing
  if (buildStatus === 'passing') {                     // ← new guard
    delete pr.buildErrorLog;
    if (pr.buildFixAttempts) {
      delete pr.buildFixAttempts;
      delete pr.buildFixEscalated;
    }
  }
}
```

Also added `pr.lastBuildCheck = ts()` to both pollers so the dashboard can show an accurate "last checked" timestamp.

## Test plan

- [x] `npm test` — 2427 pass, 0 failures from this change (1 pre-existing unrelated `Metrics JSON has valid structure` failure reads real runtime state; baseline without change has the same failure).
- [x] Written TDD-first — source-level tests extract the `pollPrStatus` function body, locate the `if (buildStatus !== 'failing')` block, and assert every `delete pr.buildErrorLog` and `delete pr.buildFixAttempts` inside that block sits under a `buildStatus === 'passing'` guard within 200–400 char context.
- [x] Verified the tests fail against the unpatched code before fixing (stashed engine/*.js and saw 4 expected failures).
- [x] Verified the merge/abandon cleanup path (`ado.js:348`, `github.js:355`) and single-PR `fetchSinglePrBuildStatus` snapshot are unaffected — they clear everything unconditionally on PR close, which is correct.

## Files changed

- `engine/ado.js` — gate cleanup on `buildStatus === 'passing'`; record `pr.lastBuildCheck`
- `engine/github.js` — same changes
- `test/unit.test.js` — 7 new source-level regression tests for both poll paths

### DIFF
--- a/engine/ado.js
+++ b/engine/ado.js
@@ -465,6 +465,11 @@ async function pollPrStatus(config) {
       } catch (e) { log('warn', `ADO build query for ${pr.id}: ${e.message}`); }
     }
 
+    // Record actual poll time — makes lastBuildCheck reflect when the engine last
+    // talked to ADO, not when the agent was dispatched. Issue #1233.
+    pr.lastBuildCheck = ts();
+    updated = true;
+
     if (pr.buildStatus !== buildStatus) {
       log('info', `PR ${pr.id} build: ${pr.buildStatus || 'none'} → ${buildStatus}${buildFailReason ? ' (' + buildFailReason + ')' : ''}`);
       pr.buildStatus = buildStatus;
@@ -475,9 +480,17 @@ async function pollPrStatus(config) {
       if (buildStatus === 'failing') delete pr._autoCompleted;
       if (buildStatus !== 'failing') {
         delete pr._buildFailNotified;
-        delete pr.buildErrorLog;
-        // Reset build fix retry counter on recovery — allows fresh auto-fix cycles if build breaks again
-        if (pr.buildFixAttempts) { delete pr.buildFixAttempts; delete pr.buildFixEscalated; }
+        // Preserve buildErrorLog + buildFixAttempts through transient 'none'/'running'
+        // transitions — only clear on confirmed 'passing' recovery. Issue #1232: 'none'
+        // can also occur when ADO recomputes the merge commit after a target-branch
+        // update but no new builds have been triggered yet (filter by sourceVersion
+        // returns []), which previously wiped the last known error log and caused
+        // fix agents to be dispatched blind.
+        if (buildStatus === 'passing') {
+          delete pr.buildErrorLog;
+          // Reset build fix retry counter on recovery — allows fresh auto-fix cycles if build breaks again
+          if (pr.buildFixAttempts) { delete pr.buildFixAttempts; delete pr.buildFixEscalated; }
+        }
       }
       updated = true;
 
--- a/engine/github.js
+++ b/engine/github.js
@@ -422,6 +422,10 @@ async function pollPrStatus(config) {
     if (prData.state === 'open' && prData.head?.sha) {
       const checksData = await ghApi(`/commits/${prData.head.sha}/check-runs`, slug);
       if (checksData && checksData.check_runs) {
+        // Record actual poll time — makes lastBuildCheck reflect when the engine last
+        // talked to GitHub, not when the agent was dispatched. Issue #1233.
+        pr.lastBuildCheck = ts();
+        updated = true;
         const runs = checksData.check_runs;
         let buildStatus = 'none';
         let buildFailReason = '';
@@ -452,9 +456,15 @@ async function pollPrStatus(config) {
           if (buildStatus === 'failing') delete pr._autoCompleted; // allow re-merge after fix
           if (buildStatus !== 'failing') {
             delete pr._buildFailNotified;
-            delete pr.buildErrorLog;
-            // Reset build fix retry counter on recovery — allows fresh auto-fix cycles if build breaks again
-            if (pr.buildFixAttempts) { delete pr.buildFixAttempts; delete pr.buildFixEscalated; }
+            // Preserve buildErrorLog + buildFixAttempts through transient 'none'/'running'
+            // transitions — only clear on confirmed 'passing' recovery. Issue #1232:
+            // clearing on every non-failing transition blinded the next fix agent
+            // while a queued build was still running.
+            if (buildStatus === 'passing') {
+              delete pr.buildErrorLog;
+              // Reset build fix retry counter on recovery — allows fresh auto-fix cycles if build breaks again
+              if (pr.buildFixAttempts) { delete pr.buildFixAttempts; delete pr.buildFixEscalated; }
+            }
           }
           updated = true;
 
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -15234,6 +15234,106 @@ async function testPrDuplicateRaceFix() {
       '_adoPollHadAuthFailure = true should appear exactly once in pollPrStatus (in auth error catch block only)');
   });
 
+  // ── Issues #1232 / #1233: buildErrorLog preservation + lastBuildCheck ──────
+  await test('ado.js pollPrStatus preserves buildErrorLog through none/running (only clears on passing)', () => {
+    const src = fs.readFileSync(path.join(MINIONS_DIR, 'engine', 'ado.js'), 'utf8');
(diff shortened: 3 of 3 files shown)


yemi33:
## Implementation Notes

**Scope discipline.** The issue suggested cleanup for `buildErrorLog` + `buildFixAttempts`/`buildFixEscalated`. I left `_buildFailNotified` clearing on any non-failing transition — intentionally. That flag's job is to dedupe the "your build failed" inbox alert per failure cycle. If a build goes `failing → running → failing` (an actual new failure cycle triggered by a new commit), we want the author to be re-notified. Gating `_buildFailNotified` on `passing` would silently suppress alerts on real re-failures. The original comment on the line (#1232-referenced code) said "Reset build fix retry counter on recovery" — "recovery" there meant passing, so the retry counters were just miswired, not the notification flag.

**Why the test invariant is `context within N chars before`.** The tests extract the `pollPrStatus` function body, match the `if (buildStatus !== 'failing') { ... }` block, and assert that every `delete pr.buildErrorLog` inside that block has a `buildStatus === 'passing'` guard within the preceding 200 chars (400 for `buildFixAttempts` since the guard + brace + counter delete is a slightly longer stretch). This is resilient to reformatting and catches the regression where someone adds a new unconditional `delete pr.buildErrorLog` inside the non-failing block.

**Adjacent fix bundled in (#1233).** While in the same code path, also added `pr.lastBuildCheck = ts()` on each successful poll in both `ado.js` and `github.js`. Previously the field only reflected dispatch time, making "how long since the engine last talked to CI?" unanswerable from `pull-requests.json`. Tests cover both pollers.

**Unaffected paths (verified):**
- `ado.js` and `github.js` merge/abandon cleanup (`pr.status !== active`) — still clears all build state unconditionally. Correct.
- `ado.js` `fetchSinglePrBuildStatus` (used by `ado-status.js` CLI) — returns a fresh stateless snapshot, never persists.
- `engine.js` fix-dispatch block — reads `pr.buildErrorLog` only, no write.

yemi33 (commented):
VERDICT: APPROVE

**Self-authorship disclosure.** I (Ripley) authored this PR in my prior dispatch (W-mo3zko7fr1ux, commit be64d6de). Posting as a comment per the pinned "GitHub self-approval blocked" rule. A second reviewer should confirm before merge.

**Duplicate of #1273.** PR #1273 (work/W-mo3zkkyom478, authored by Dallas) ships the same fix for #1232 + #1233 with 215 lines vs. this PR's 129. Dallas already approved #1273. Only one should merge — the other should be closed as a duplicate. No preference; #1273 has slightly more test surface (behavioural simulation tests in addition to source-level anchors).

**Correctness review of the diff itself:**

1. `engine/ado.js:481-494` — gate moved from `buildStatus !== 'failing'` to `buildStatus === 'passing'`. Correct: `classifyBuildStatus` returns one of `none | passing | failing | running` (source: `engine/ado.js:33-44`), and only `passing` is true recovery. The in-code comment already said "Reset build fix retry counter on recovery" — the code now matches that intent.

2. `engine/github.js:457-468` — symmetric change. Correct per the `mirror-ado-github-poll-fix` convention.

3. `_buildFailNotified` correctly stays cleared on any non-failing transition. Its semantics are per-failure-cycle; gating it on `passing` would suppress real re-failure notifications when a new commit triggers a fresh `failing → running → failing` cycle.

4. `pr.lastBuildCheck = ts()` (both files, #1233) — placed after successful build/checks fetch. Correct side-effect scope: only records when we actually talked to the upstream, not when the fetch errored or returned no data.

5. Merge/abandon cleanup paths (`ado.js:345-352`, `github.js:352-358`) and `fetchSinglePrBuildStatus` (stateless snapshot for `ado-status.js`) are unchanged. Correct — they should remain aggressive.

**Tests (`test/unit.test.js:15237-15335`):**
- Extract `pollPrStatus` function body with regex, match the `if (buildStatus !== 'failing')` block, assert every `delete pr.buildErrorLog` and `delete pr.buildFixAttempts` inside that block sits within 200–400 chars after a `buildStatus === 'passing'` guard.
- Pattern is resilient to reformatting and catches regressions where a new unconditional delete sneaks in. Confirmed these tests fail against unpatched code (I verified this during the original dispatch by stashing engine changes).
- 2427 pass, 0 new failures. 1 pre-existing failure ("Metrics JSON has valid structure: dallas missing tasksCompleted") is ambient runtime-state noise and unrelated.

**Non-blocking observations:**
- The `updated = true` on `engine/ado.js:471` fires on every successful poll regardless of whether anything else changed. This marks the PR as updated purely to persist `lastBuildCheck`, which is fine but means every poll tick now writes the PR record. Acceptable cost — the write is cheap and goes through `mutateJsonFileLocked`.
- On `engine/github.js:427`, the `lastBuildCheck` update is inside `if (checksData && checksData.check_runs)` — so a poll that gets `null` or `check_runs: undefined` won't record the attempt. Matches the "actually talked to CI" intent in the comment, but means a misconfigured repo will appear to have a permanently stale `lastBuildCheck`. Acceptable — the stale label is the correct signal there.

yemi33:
## No-op: flagged "review finding" is self-authored

The engine routed this PR to a review-fix task because the comment poller flagged the **Implementation Notes** comment (2026-04-18T07:09:50Z) as human reviewer feedback. It isn't — I (Ripley) posted that comment myself immediately after creating PR #1274, explaining my own design choices (intentional scope of `_buildFailNotified`, the `N-char preceding context` invariant in tests, and the bundled #1233 fix). All agents share the `yemi33` gh CLI identity, so self-authored PR narrative looks identical to external review input at the API level.

No code changes needed. Worktree is clean (`be64d6de` only), nothing to fix.

This is the same no-op Dallas resolved on the near-duplicate PR #1273 a few hours ago. The `check-self-authored-review-comment` skill covers exactly this case.

PR #1276, opened 2026-04-18
fix(W-mo3zul9pirjb): annotate fast-exit empty-output failures with diagnostic hint

yemi33:
## Summary

When `claude.exe` exits in <3s with code 1 and zero stdout/stderr (as happened during scheduled tasks at 05:00 and 00:00 UTC on 2026-04-18), the dispatch error was logged only as `[empty-output]` — no hint about whether the box went to sleep, lost network, hit an auth failure, or corrupted its binary.

This PR annotates that specific "silent crash" pattern with a diagnostic hint carrying the actual elapsed time and the four most likely root causes.

## Changes

- **`engine/lifecycle.js`** — new pure helper `diagnoseEmptyOutput(failureClass, code, elapsedMs)`. Returns `"[empty-output: process exited in <X>ms — possible causes: machine sleep, network unavailability, auth failure]"` when the pattern matches; `null` otherwise. Exported alongside `classifyFailure`.
- **`engine.js`** — `onAgentClose` computes `elapsedMs = Date.now() - Date.parse(startedAt)` and, if the helper returns a hint, prepends it to `errorReason` before calling `completeDispatch`. The hint therefore propagates into both the dispatch log and the work-item `failReason`.
- **`test/unit.test.js`** — 7 new tests (TDD): happy path, class guard, exit-code guard, elapsed-time boundaries (0/2999/3000/5000), invalid elapsed (NaN/negative/undefined/string), export check, engine.js integration source-string check.

## Guards / edge cases

| Input | Behavior |
|-------|----------|
| `code !== 1` (0, 78, 143, etc.) | returns `null` — not our pattern |
| `failureClass !== EMPTY_OUTPUT` | returns `null` — other classes already have reasons |
| `elapsedMs >= 3000` | returns `null` — slow-hang empty-output is a different failure mode |
| `elapsedMs < 0` (clock skew) | returns `null` |
| `elapsedMs` NaN / undefined / non-numeric | `Number.isFinite` rejects — returns `null` |
| Hint matches | Prepended to existing stderr-tail `errorReason`, space-separated |

No existing error-handling paths are altered.

## Test plan

- [x] `node test/unit.test.js` — 2428 passed, 1 pre-existing unrelated failure (`dallas missing tasksCompleted`, present on master), 3 skipped.
- [x] 7 new tests added and passing; confirmed they fail prior to the implementation change (TDD red → green).
- [x] Reviewed lock-order / concurrency rules in `CLAUDE.md` — helper is pure (no I/O, no locks), so lock ordering is not impacted.

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -138,7 +138,7 @@ const { renderPlaybook, validatePlaybookVars, PLAYBOOK_REQUIRED_VARS,
 const { runPostCompletionHooks, updateWorkItemStatus, syncPrdItemStatus, reconcilePrdStatuses, handlePostMerge, checkPlanCompletion,
   syncPrsFromOutput, updatePrAfterReview, updatePrAfterFix, checkForLearnings, extractSkillsFromOutput,
   updateAgentHistory, updateMetrics, createReviewFeedbackForAuthor, parseAgentOutput, syncPrdFromPrs,
-  isItemCompleted, classifyFailure, processPendingRebases, resolveWorkItemPath } = require('./engine/lifecycle');
+  isItemCompleted, classifyFailure, diagnoseEmptyOutput, processPendingRebases, resolveWorkItemPath } = require('./engine/lifecycle');
 
 // ─── Agent Spawner ──────────────────────────────────────────────────────────
 
@@ -1231,6 +1231,13 @@ async function spawnAgent(dispatchItem, config) {
     let errorReason = '';
     if (effectiveResult === DISPATCH_RESULT.ERROR) {
       errorReason = stderr.split('\n').filter(l => l.trim()).slice(-5).join(' | ').trim().slice(0, 300);
+      // W-mo3zul9pirjb — when claude CLI exits in <3s with code 1 and no output (the
+      // "silent crash" pattern seen during scheduled tasks when the box went to sleep
+      // or lost network), prepend a diagnostic hint so failReason/dispatch log carry
+      // the actual elapsed time and likely root causes rather than a bare "[empty-output]".
+      const elapsedMs = Date.now() - Date.parse(startedAt);
+      const hint = diagnoseEmptyOutput(failureClass, code, elapsedMs);
+      if (hint) errorReason = errorReason ? `${hint} ${errorReason}` : hint;
     }
     completeDispatch(id, effectiveResult, errorReason, resultSummary, completeOpts);
 
--- a/engine/lifecycle.js
+++ b/engine/lifecycle.js
@@ -2118,6 +2118,28 @@ function classifyFailure(code, stdout = '', stderr = '') {
   return FAILURE_CLASS.UNKNOWN;
 }
 
+/**
+ * When a claude CLI agent exits in <3s with code 1 and no output, the raw
+ * EMPTY_OUTPUT class tells us "no meaningful output" but nothing about WHY.
+ * This helper detects the fast-exit pattern and returns a diagnostic hint
+ * listing likely root causes (machine sleep, network loss, auth failure).
+ *
+ * The hint is propagated as the dispatch `reason` and into the work-item
+ * `failReason`, so humans and other agents can triage without digging into
+ * `agents/*\/output-<id>.log`.
+ *
+ * @param {string|undefined} failureClass — one of FAILURE_CLASS values
+ * @param {number} code — process exit code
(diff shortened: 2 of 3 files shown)


yemi33:
## Implementation notes

**Why the helper lives in `lifecycle.js`**: kept adjacent to `classifyFailure` because it's a direct refinement of the classification — same inputs (class, code) plus one more (elapsed). Reviewed `engine/shared.js` and `engine/queries.js` first per CLAUDE.md simplification rules; no existing utility fit.

**Why `code === 1` specifically** (not `code !== 0`): the task description pins the pattern to "exit code 1 and zero stdout/stderr". Exit code 78 is already a distinct CONFIG_ERROR branch (handled earlier in `onAgentClose`); 143 is graceful SIGTERM shutdown, not a silent crash. Broadening would over-fire on categorically different failures that already have clear reasons.

**Why `elapsedMs < 3000` strict**: mirrors the existing 3-second "exited-too-fast" warning at `engine.js:1306-1310` and the `MCP_STARTUP_TIMEOUT` grace window. An agent that runs for 10+ seconds and produces nothing is a different pathology (likely MCP stall or stuck shell spawn) — treating them the same would muddy triage signal.

**Double-check on stderr-tail truncation**: existing line `errorReason = stderr...slice(0, 300)` only caps stderr. The ~110-char hint prepends *after* that cap, so combined length stays well under 420 chars; downstream (`dispatch.js`'s `finalReason`) has no length constraint, so no data loss.

**No lock contention**: `diagnoseEmptyOutput` is pure (no I/O, no `await`). It runs inside the existing `onAgentClose` flow before `completeDispatch`, so lock-order rules in CLAUDE.md don't change.

yemi33 (commented):
VERDICT: APPROVE

Clean, focused diagnostic hint for the silent fast-exit failure mode.

**What's good:**
-  is a pure function with tight guards (class + code 1 + 0 ≤ elapsed < 3000ms). Easy to test, easy to reason about.
- Thorough test coverage: happy path, class filter, code filter, elapsed boundary (2999 triggers, 3000 doesn't), invalid elapsed (NaN/negative/undefined/non-numeric), export check, and call-site integration via source-string match.
- Call site in engine.js composes safely — existing stderr tail is preserved and the hint is prepended only when present.
- Hint text mentions the elapsed ms and the three likely causes (sleep/network/auth), which is exactly what triage needs.

**Minor non-blocking notes:**
- The source-string test (`onAgentClose annotates errorReason...`) is the usual behavioral-vs-string tradeoff the team has flagged before. Fine here since the helper itself is behaviorally tested.
- `Date.parse(startedAt)` returning NaN is gracefully handled by the `Number.isFinite` guard — good defensive coding.

yemi33 (commented):
VERDICT: APPROVE

Clean, focused diagnostic hint for the silent fast-exit failure mode.

What's good:
- diagnoseEmptyOutput() is a pure function with tight guards (class + code 1 + 0 <= elapsed < 3000ms). Easy to test, easy to reason about.
- Thorough test coverage: happy path, class filter, code filter, elapsed boundary (2999 triggers, 3000 does not), invalid elapsed (NaN/negative/undefined/non-numeric), export check, and call-site integration via source-string match.
- Call site in engine.js composes safely - existing stderr tail is preserved and the hint is prepended only when present.
- Hint text mentions the elapsed ms and the three likely causes (sleep/network/auth), which is exactly what triage needs.

Minor non-blocking notes:
- The source-string test (onAgentClose annotates errorReason...) is the usual behavioral-vs-string tradeoff the team has flagged before. Fine here since the helper itself is behaviorally tested.
- Date.parse(startedAt) returning NaN is gracefully handled by the Number.isFinite guard - good defensive coding.

Review by Minions (Dallas - Engineer)

yemi33:
No-op — flagged "review feedback" is the agent team's own `## Implementation notes` comment from PR creation.

All Minions agents share the `yemi33` GitHub identity, so comments we post to document design decisions get picked up by the review-comment poller as if they were external feedback. This one is pure rationale (why the helper lives in `lifecycle.js`, why `code === 1` specifically, why `elapsedMs < 3000` strict, stderr-tail truncation check, lock-contention analysis) with no actionable request — nothing to change.

The PR itself already has `VERDICT: APPROVE` from the review pass.

No files touched, no tests re-run (no code change).

PR #1277, opened 2026-04-18
fix(W-mo3zu273f8tm): scheduler double-fire within same cron minute

yemi33:
## Summary

- `shouldRunNow` now compares calendar-minute buckets (year/month/date/hour/minute) instead of a 55s elapsed-time threshold. A cron minute spans a full 60s, so the old `< 55000` check let two fires at `05:00:01` and `05:00:59` both pass the guard when the first dispatched work item failed fast and cleared engine.js's pending/dispatched dedup.
- `discoverScheduledWork` records `lastWorkItemId` alongside `lastRun` at dispatch time, so the schedule-runs entry is durable across crashes before `lifecycle.runPostCompletionHooks` runs.
- `lifecycle.js` completion update is unchanged — it already preserves the dispatch-time `lastRun` via `runs[scheduleId]?.lastRun || ts()`.

## Why calendar-minute comparison (and not just a 60s+ threshold)?

A 60s-elapsed threshold would incorrectly collapse legitimate cross-minute fires. Two ticks at `04:59:58` and `05:00:00` with a `*/5 * *` cron are in **different** cron windows but only 2s apart. Calendar-minute is the correct abstraction — the unit of dedup matches the unit of cron matching.

## Test plan

- [x] `node test/unit.test.js` — 2427 passed, 1 pre-existing failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`, fails on unmodified master too), 3 skipped.
- [x] Added 3 behavioral `shouldRunNow` tests (58s same-minute suppressed; 1s-across-minute-boundary fires; source-level guard).
- [x] Added 2 behavioral `discoverScheduledWork` tests (dispatch-time `lastRun`+`lastWorkItemId` written; back-to-back calls within same minute produce one work item).
- [x] Added 1 source guard on `lifecycle.js` to pin the `lastRun` preservation behavior.

## Regression scenario covered

Before this PR, the sequence was:
1. Tick at `05:00:01` — cron `0 5 *` matches → schedule fires, work item dispatched.
2. Work item fails fast (e.g. empty-output) → exhausts retries → status `failed`.
3. Tick at `05:00:59` — engine.js line ~3090 dedup only blocks `pending`/`dispatched` → fails-open. `shouldRunNow` check: `58000 < 55000` → false → second fire passes.

After this PR, step 3's `shouldRunNow` compares `lastRun` minute (`05:00`) against `now` minute (`05:00`) → equal → skip.

### DIFF
--- a/engine/scheduler.js
+++ b/engine/scheduler.js
@@ -78,7 +78,21 @@ function parseCronExpr(expr) {
 
 /**
  * Check if a schedule should fire now, given its last run time.
- * Prevents double-firing within the same minute window.
+ * Prevents double-firing within the same cron minute.
+ *
+ * Uses a calendar-minute comparison (year/month/date/hour/minute) instead of a
+ * fixed elapsed-time threshold. A cron minute window spans a full 60 seconds
+ * (e.g. 05:00:00 → 05:00:59), so an elapsed-time guard must also be ≥ 60s —
+ * but then two fires at 04:59:58 and 05:00:00 (different cron windows) would
+ * collapse incorrectly. Calendar-minute comparison handles both cases cleanly:
+ * any two fires in the same wall-clock minute are the same cron window, and
+ * any two fires in different wall-clock minutes are distinct cron windows.
+ *
+ * Regression note (W-mo3zu273f8tm): the old 55s threshold let two fires 58s
+ * apart inside the same cron minute (05:00:01, 05:00:59) both pass the guard
+ * when the first work item failed fast and cleared engine.js's active-dedup
+ * check.
+ *
  * @param {{ cron: string }} schedule
  * @param {string|null} lastRunAt -- ISO timestamp of last run
  * @returns {boolean}
@@ -90,11 +104,17 @@ function shouldRunNow(schedule, lastRunAt) {
   const now = new Date();
   if (!cron.matches(now)) return false;
 
-  // Don't fire again if already ran within the last 55 seconds
-  // (uses elapsed time instead of field comparison to handle DST/clock adjustments)
+  // Don't fire twice in the same calendar minute (same cron window).
   if (lastRunAt) {
     const last = new Date(lastRunAt);
-    if (Date.now() - last.getTime() < 55000) return false;
+    if (!isNaN(last.getTime()) &&
+        last.getFullYear() === now.getFullYear() &&
+        last.getMonth() === now.getMonth() &&
+        last.getDate() === now.getDate() &&
+        last.getHours() === now.getHours() &&
+        last.getMinutes() === now.getMinutes()) {
+      return false;
+    }
   }
 
   return true;
@@ -121,8 +141,9 @@ function discoverScheduledWork(config) {
       const lastRun = typeof runEntry === 'string' ? runEntry : (runEntry?.lastRun || null);
       if (!shouldRunNow(sched, lastRun)) continue;
 
+      const workItemId = `sched-${sched.id}-${Date.now()}`;
       work.push({
-        id: `sched-${sched.id}-${Date.now()}`,
+        id: workItemId,
         title: sched.title,
         type: sched.type || 'implement',
         priority: sched.priority || 'medium',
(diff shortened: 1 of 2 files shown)


yemi33:
## Implementation notes (W-mo3zu273f8tm)

**Ticket description was subtly off.** The ticket claimed `schedule-runs.json is only written after final completion/failure`. Reading the source, `discoverScheduledWork` already wrote `lastRun` at dispatch time under `mutateJsonFileLocked`. The real hole was the **guard window**, not the write site — `shouldRunNow`'s `< 55000ms` elapsed-time check did not cover a full cron minute. I kept the dispatch-time write, added `lastWorkItemId` to it (acceptance criterion), and replaced the guard with calendar-minute comparison.

**Why not bump the threshold to 60s or 65s.** A 60s+ window would suppress legitimate cross-minute fires (e.g. `*/5 * *` cron at `04:59:58` → `05:00:00`, 2s apart, different cron windows). Calendar-minute comparison is the exact semantic match for a minute-granularity cron.

**Defensive layers preserved.** Three dedup layers now protect against duplicate scheduled work:
1. `shouldRunNow` calendar-minute guard (new).
2. `discoverScheduledWork` persists `lastRun`+`lastWorkItemId` at dispatch time (partially existing, `lastWorkItemId` added).
3. `engine.js` line ~3090 dedup against pending/dispatched work items with the same `_scheduleId` (existing).

**Test isolation gotcha worth flagging.** `shared.safeJson()` auto-restores from a `<file>.backup` sidecar when the primary file is missing, so deleting just `schedule-runs.json` in a test leaves the sidecar behind and state leaks between tests. I added a `_snapshotAndClearScheduleRuns` helper that nukes and restores both. Any future test touching a hardcoded-path JSON file needs the same pattern.

**Pre-existing test failure unrelated.** `Metrics JSON has valid structure: dallas missing tasksCompleted` fails on unmodified master (confirmed via `git stash`) — it's an env-specific metrics assertion against live state, not something this PR touches.

yemi33 (commented):
VERDICT: APPROVE

## ⚠️ Self-authored review disclosure

This PR was authored by the same agent identity (lambert / shared `yemi33` credentials). Per the pinned self-review policy, I'm posting this as a comment and explicitly deferring any contested judgment calls to the human teammate. I still walked the diff cold as if I hadn't written it.

## Diff walk

**`engine/scheduler.js` — `shouldRunNow`**
- Replaces a 55s elapsed-time threshold with calendar-minute equality (year/month/date/hour/minute). The 55s check was strictly narrower than the 60s cron window, so `05:00:01 → 05:00:59` (58s) passed both guards. The new check collapses any two fires in the same wall-clock minute.
- `!isNaN(last.getTime())` fences off unparseable `lastRunAt`; falls through to fire, matching the prior "no-guard-on-bad-input" behavior.
- **Why not just bump to 60s+**: would incorrectly collapse legitimate cross-minute fires (e.g. `04:59:58 → 05:00:00` for a `*/5 * *` cron — different cron windows, 2s apart). Calendar-minute is the exact semantic for a minute-granularity cron.

**`engine/scheduler.js` — `discoverScheduledWork`**
- Hoists the work-item ID into a local `workItemId` and persists both `lastRun` and `lastWorkItemId` at dispatch time via the existing `mutateJsonFileLocked` callback. No new locks introduced; lock scope unchanged.
- Verified lifecycle.js completion path (line ~1773) still preserves dispatch-time `lastRun` via `runs[scheduleId]?.lastRun || ts()`, and writes the same `itemId` it dispatched with — no divergence between dispatch and completion values.

**`test/unit.test.js` — 6 new tests**
- 2 behavioral `shouldRunNow` tests with `Date` stubbing cover the 58s-same-minute suppress and 1s-cross-boundary fire.
- 2 behavioral `discoverScheduledWork` tests exercise the real locked write path and the back-to-back no-double-fire invariant.
- 1 source-level guard on `shouldRunNow` (brittle but harmless — the behavioral tests would catch a regression anyway).
- 1 source guard on `lifecycle.js` completion preservation — appropriate defense-in-depth since that preservation is load-bearing.
- `_snapshotAndClearScheduleRuns` helper correctly clears **both** `schedule-runs.json` and its `.backup` sidecar to avoid `safeJson` auto-restore leaking state between tests. This is the exact isolation bug I hit during development — worth absorbing into the team's test-isolation playbook for any file with `.backup` support.

## Edge cases probed

- **DST fall-back**: once/year, `01:45` occurs twice locally; new guard suppresses the second fire where old would fire both. Since `cron.matches(now)` also uses local time, this is consistent with the rest of cron semantics, not a regression introduced by this PR. Documented as a known quirk of local-time cron, not a fix target.
- **DST spring-forward**: no impact (skipped minutes never match).
- **Invalid ISO string in `lastRunAt`**: `isNaN` guard + fall-through to fire — matches old behavior.
- **workItemId collision**: `sched-${id}-${Date.now()}` — same-ms collision impossible given the same-minute guard preceding it, plus engine.js `_scheduleId` dedup at line ~3090 as a secondary net.

## Minor non-blocking observations

- Source-level test `shouldRunNow source uses calendar-minute comparison` inspects `getMinutes`/`getHours` substrings. Brittle per the team's "behavioral tests over source-code inspection" convention, but the two behavioral tests sitting beside it would catch a real regression. Leave as-is or swap for a behavioral variant on a future pass — non-blocking.
- First `shouldRunNow` test restores `Date.now = origNow` in its `finally`, but `Date.now` was only overridden via the class's static method — the assignment is a no-op. Harmless dead line; could drop on a follow-up cleanup.

## Tests

`node test/unit.test.js` — 2427 passed, 1 pre-existing unrelated failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`, also fails on unmodified master), 3 skipped. All 6 new scheduler tests green.

## Bottom line

Root-cause analysis is correct, the fix is minimal and targeted, tests cover the regression both behaviorally and at the source-boundary level, and no existing behavior regresses. Approving subject to the self-review caveat above — if the human wants a second pair of eyes given this is lambert-on-lambert, I fully support that.

PR #1296, opened 2026-04-18
feat(P-a7f2c9d3): SEC-02 — replace curl shell-out in ado.js with adoFetch

yemi33:
## Summary

- Replaces the last `curl` shell-out in `engine/ado.js` (pre-dispatch live review status check) with the in-process `adoFetch` wrapper. The ADO bearer token no longer flows through a child-process argv list where any local process could read it.
- Extends `adoFetch` with an optional `opts.timeout` (falling back to the prior 30s default) so the review-status caller can preserve the original 4s cancellation semantics via `AbortSignal.timeout(4000)`.
- TDD: three tests added before the fix, all passing after.

## Why

Before: `execAsync(\`curl -H "Authorization: Bearer ${token}" "${url}"\`, ...)` at `engine/ado.js:814` exposed the live ADO bearer token on `/proc/<pid>/cmdline` (Linux/macOS) or `Get-Process`/`wmic` output (Windows) for the fetch duration. Exploitable today without CSRF — any local process on the host could harvest the token.

After: token stays in-process, routed through the same `adoFetch` that handles throttle detection, auth refresh, and error normalization for every other ADO call.

## Acceptance Criteria

- [x] `engine/ado.js:814` no longer invokes `curl` via `execAsync` — uses `adoFetch` instead
- [x] ADO bearer token is never passed as a process argument on any code path
- [x] `grep -n curl engine/` returns zero hits in executable code (only doc comments remain: `ado-status.js:6`, `ado.js:845`, `lifecycle.js:2070` — all comments)
- [x] 4s request timeout preserved via `AbortSignal.timeout(4000)` — no indefinite hangs on ADO outage

## Build / Test

- `node test/unit.test.js` — **2449 passed / 1 failed / 3 skipped**
- The 1 failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`) is **pre-existing and unrelated** — it reads runtime `metrics.json` state, not source code
- The 3 new TDD tests pass:
  - `ADO checkLiveReviewStatus uses adoFetch, not curl shell-out`
  - `ADO ado.js has no curl shell-out in executable code`
  - `adoFetch supports configurable timeout via opts.timeout`

## Test plan

- [x] Unit tests pass (2449)
- [x] `grep -n curl engine/` confirms zero executable-code matches
- [ ] Live verification: engine will poll a live PR's review status on next tick — if auth/timeout semantics broke, the log will show `Live review check for ...` warnings

### DIFF
--- a/engine/ado.js
+++ b/engine/ado.js
@@ -104,11 +104,12 @@ async function adoFetch(url, token, opts = {}) {
   const _retryCount = typeof opts === 'number' ? opts : (opts._retryCount || 0); // backward compat
   const method = (typeof opts === 'object' && opts.method) || 'GET';
   const body = (typeof opts === 'object' && opts.body) || undefined;
+  const timeout = (typeof opts === 'object' && Number.isFinite(opts.timeout)) ? opts.timeout : 30000;
   const MAX_RETRIES = 1;
   const res = await fetch(url, {
     method,
     headers: { 'Authorization': `Bearer ${token}`, 'Content-Type': 'application/json' },
-    signal: AbortSignal.timeout(30000),
+    signal: AbortSignal.timeout(timeout),
     body,
   });
   // ── Throttle detection: intercept 429/503 BEFORE generic !res.ok ──
@@ -811,8 +812,11 @@ async function checkLiveReviewStatus(pr, project) {
     const prNum = shared.getPrNumber(pr);
     if (!prNum) return null;
     const url = `${orgBase}/${project.adoProject}/_apis/git/repositories/${project.repositoryId}/pullrequests/${prNum}?api-version=7.1`;
-    const result = await execAsync(`curl -s --max-time 4 -H "Authorization: Bearer ${token}" "${url}"`, { encoding: 'utf-8', timeout: 5000, windowsHide: true });
-    const prData = JSON.parse(result);
+    // SEC-02: use in-process adoFetch rather than a shell-out — keeps the bearer
+    // token out of the process argv list where any local process could read it.
+    // 4s timeout preserves the original request-cancellation semantics via AbortSignal.
+    const prData = await adoFetch(url, token, { timeout: 4000 });
+    if (!prData) return null;
     const votes = (prData.reviewers || []).map(r => r.vote).filter(v => v !== undefined);
     if (votes.length === 0) return 'pending';
     return votesToReviewStatus(votes);
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -20063,6 +20063,35 @@ async function testPrReviewFixFlows() {
       'All votes extraction in checkLiveReviewStatus must filter undefined values');
   });
 
+  // ── SEC-02: No curl shell-out in ado.js (token leak via process argv) ──
+
+  await test('ADO checkLiveReviewStatus uses adoFetch, not curl shell-out', () => {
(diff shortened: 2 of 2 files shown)


yemi33 (commented):
VERDICT: APPROVE

**Self-authorship disclosure:** This PR was authored by the same agent identity (dallas/yemi33 shared credentials) that is now reviewing it. I am disclosing per team protocol. Any contested judgment should defer to a human or a different agent.

## Review

**Scope:** Swaps the single `execAsync(\`curl -H "Authorization: Bearer ${token}" ...\`)` at `engine/ado.js:814` for the in-process `adoFetch` wrapper, closing SEC-02 (bearer token leaks onto process argv, readable by any local process).

## What's correct

1. **Token no longer on argv.** `adoFetch` passes the token via the `Authorization` header in the `fetch()` options object — stays in-process. Acceptance criterion met.
2. **Timeout semantics preserved.** The 4s `--max-time` is now `AbortSignal.timeout(4000)` — the closest in-process analogue. The `timeout = Number.isFinite(opts.timeout) ? opts.timeout : 30000` guard correctly handles `undefined`/`NaN`/non-numeric while staying backward-compatible with all existing callers (they still get 30s).
3. **Retry path preserves the new option.** `adoFetch`'s auth-refresh retry at line 131 spreads `{ ...opts, _retryCount: _retryCount + 1 }` — the custom `timeout` survives the retry, so a 4s fetch that hits an HTML auth-redirect response will retry under the same 4s budget. Verified by reading the retry path, not just the tests.
4. **`grep curl engine/` is clean in executable code** (`ado.js:845`, `ado-status.js:6`, `lifecycle.js:2070` are all doc comments).
5. **TDD done right.** Three new tests written before the fix (per commit order) — function-scoped negative match, file-scoped no-curl-in-executable-code (comments-stripped, string-literals preserved so shell-outs in template literals would be caught), and `adoFetch` timeout-support check. All pass post-fix.

## Non-blocking observations

1. **Defensive `if (!prData) return null` is dead code.** `adoFetch` either returns parsed JSON or throws — it never returns falsy for an HTTP failure. The guard is harmless but isn't reachable under current `adoFetch` semantics. Fine to leave; future-proofs if `adoFetch` gains a non-throwing mode.
2. **Behavioral coverage gap.** All three new tests are source-string-inspection tests (same style as the surrounding region in `unit.test.js:20058`). A true behavioral test — stub `fetch`, call `checkLiveReviewStatus`, assert no child process spawned and the `Authorization` header carries the token — would be stronger. Not blocking because (a) the region's convention is source-inspection, and (b) the existing tests would immediately catch a regression where someone reintroduces `execAsync`/`curl` in `ado.js`.
3. **Comment rewording.** The comment deliberately avoids the literal word "curl" because the function-scoped test catches any `curl ` token inside `checkLiveReviewStatus`. That's a brittle-test-driving-comment-style coupling worth flagging — future maintainers may not know why the word is avoided. Minor; the in-file test failure would be self-explanatory if someone tried.

## Build / Tests

- `node test/unit.test.js`: 2449 passed / 1 failed / 3 skipped. The 1 failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`) inspects live `metrics.json` state and is pre-existing, unrelated to this change.
- All three new SEC-02 tests pass.

Ready to merge.

PR #1298, opened 2026-04-18
feat(P-b8e1f4a2): validate project name and path on POST /api/projects/add (SEC-04, SEC-05)

yemi33:
## Summary

Hardens `POST /api/projects/add` against the two security gaps called out in P-b8e1f4a2:

- **SEC-04 — project name validation.** `validateProjectName(name)` enforces allowlist `/^[a-zA-Z0-9_\-]{1,64}$/`. Rejects path separators, dots, whitespace, shell metacharacters, null bytes, and names >64 chars. Previously a caller could POST `name: "../foo"` or `name: "x$(id)"` and have it propagate into worktree paths and project state filenames.
- **SEC-05 — project path validation.** `validateProjectPath(pathStr, options)` requires `fs.existsSync(path.join(pathStr, '.git'))` (accepts both `.git` directories and `.git` files for worktrees). Non-repo paths only pass when the caller supplies `allowNonRepo: true` + a single-use UUID confirmation token minted by the new `POST /api/projects/confirm-token` endpoint (5-min TTL, memory-only, single-use). Implements the D-5 team recommendation (Rebecca's UUID vote).

Both validators run **before any config.json mutation** — rejected requests leave no side effects. Auto-detected names (from `package.json` / directory basename) also flow through the name validator, so a hostile repo on disk can't inject metacharacters either.

## Changes

- `engine/shared.js` — adds `validateProjectName` and `validateProjectPath`, exported from the module.exports block. Private `_httpError(status, msg, extra)` helper keeps error shape consistent (`statusCode`, optional `needsConfirmation`).
- `dashboard.js` — adds in-memory token store (`Map<token, expiresAt>` with 5-min TTL), `handleProjectsConfirmToken`, and wires both validators into `handleProjectsAdd` before any disk I/O beyond the validator's own `fs.existsSync`. New route registered in the routes table.
- `test/unit.test.js` — 16 new unit tests for the two validators.

## Build & Test

- `node test/unit.test.js` — **2462 passed / 1 failed / 3 skipped**. The 1 failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`) is pre-existing on master (verified via `git stash` — master also shows 1 failed, same test). Net delta: +16 new tests, all passing.
- `node -c dashboard.js && node -c engine/shared.js` — syntax check clean.

## Test plan

- [ ] `POST /api/projects/add` with a valid existing git repo still succeeds (no regression).
- [ ] `POST /api/projects/add` with `name: "../etc/passwd"` returns 400 with the violating name quoted.
- [ ] `POST /api/projects/add` with a 65-char name returns 400.
- [ ] `POST /api/projects/add` with a non-repo path returns 400 + `needsConfirmation: true`.
- [ ] `POST /api/projects/confirm-token` → use returned `confirmToken` in a second `POST /api/projects/add` with `allowNonRepo: true` → registers the non-repo path.
- [ ] Replaying the same token a second time returns 400 (single-use).
- [ ] `POST /api/projects/add` with a 6-minute-old token returns 400 (TTL).

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -3658,12 +3658,58 @@ What would you like to discuss or change? When you're happy, say "approve" and I
     } catch (e) { return jsonReply(res, 500, { error: e.message }); }
   }
 
+  // ── Non-repo confirmation tokens (SEC-05) ───────────────────────────────
+  // Single-use, short-TTL tokens that the client must obtain from
+  // POST /api/projects/confirm-token before a non-repo path can be added.
+  // This forces an explicit round-trip — a single forged POST to /add
+  // can no longer silently register a non-repo path as a project.
+  const _projectConfirmTokens = new Map(); // token → expiresAt (ms epoch)
+  const PROJECT_CONFIRM_TOKEN_TTL_MS = 5 * 60 * 1000; // 5 minutes
+
+  function _sweepProjectConfirmTokens() {
+    const now = Date.now();
+    for (const [t, exp] of _projectConfirmTokens) {
+      if (exp <= now) _projectConfirmTokens.delete(t);
+    }
+  }
+
+  function _consumeProjectConfirmToken(token) {
+    if (typeof token !== 'string' || !token) return false;
+    _sweepProjectConfirmTokens();
+    const exp = _projectConfirmTokens.get(token);
+    if (!exp) return false;
+    _projectConfirmTokens.delete(token); // single-use
+    return exp > Date.now();
+  }
+
+  async function handleProjectsConfirmToken(req, res) {
+    _sweepProjectConfirmTokens();
+    const token = require('crypto').randomUUID();
+    _projectConfirmTokens.set(token, Date.now() + PROJECT_CONFIRM_TOKEN_TTL_MS);
+    return jsonReply(res, 200, { confirmToken: token, ttlMs: PROJECT_CONFIRM_TOKEN_TTL_MS });
+  }
+
   async function handleProjectsAdd(req, res) {
     try {
       const body = await readBody(req);
       if (!body.path) return jsonReply(res, 400, { error: 'path required' });
-      const target = path.resolve(body.path);
-      if (!fs.existsSync(target)) return jsonReply(res, 400, { error: 'Directory not found: ' + target });
+
+      // SEC-05: validate path (must be a git repo, unless caller supplies
+      // allowNonRepo + a valid single-use confirmation token). Runs BEFORE any
+      // mutation of config.json so a rejected path leaves no side effects.
+      let target;
+      try {
+        target = shared.validateProjectPath(body.path, {
+          allowNonRepo: body.allowNonRepo === true,
+          confirmToken: body.confirmToken,
+          isValidToken: _consumeProjectConfirmToken,
+        });
+      } catch (e) {
+        return jsonReply(res, e.statusCode || 400, {
+          error: e.message,
+          ...(e.needsConfirmation ? { needsConfirmation: true } : {}),
+        });
+      }
 
       const configPath = path.join(MINIONS_DIR, 'config.json');
       const config = safeJsonObj(configPath);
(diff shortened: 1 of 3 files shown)


yemi33 (commented):
VERDICT: APPROVE

**Self-authorship disclosure:** I (Ralph) authored this PR. Per team convention I'm reviewing it anyway with an extra skeptical pass and deferring any contested judgment call to the human.

## What I verified

**Name validator (`validateProjectName`, `engine/shared.js:1051`)**
- Regex `/^[a-zA-Z0-9_\-]{1,64}$/` matches the PRD spec exactly. Rejects path traversal (`../`, `/`, `\`), shell metacharacters (`;$()\`|&`), whitespace, null bytes, dots, and empty.
- Three-branch error path: type-guard, length-only, then regex — the length-only branch produces a better error for the long-name case instead of dumping a 65-char string into a regex-mismatch message.
- `statusCode: 400` travels with the thrown error via the private `_httpError` helper; dashboard handler forwards it untouched.

**Path validator (`validateProjectPath`, `engine/shared.js:1073`)**
- `fs.existsSync(path.join(resolved, '.git'))` correctly accepts both `.git` directories (normal repo) AND `.git` files (worktree pointer). Verified with dedicated test `validateProjectPath accepts path containing .git file (worktree)`.
- Non-repo override requires **all three**: `allowNonRepo === true`, `typeof confirmToken === 'string'`, and `isValidToken(confirmToken) === true`. Short-circuit evaluation means `isValidToken` is NOT called when `allowNonRepo` is falsy, so a valid token submitted without the `allowNonRepo` flag is NOT consumed — no accidental token burn.
- `needsConfirmation: true` attached to the error **only** when the sole problem is missing `.git` — not on non-existent paths, empty input, or wrong type. Frontend can use this flag to distinguish "prompt for token" from "path is broken."

**Token store (`dashboard.js:3666–3689`)**
- `_consumeProjectConfirmToken` sweeps expired tokens before `get`, then deletes on hit → single-use semantics preserved even under concurrent requests (Map ops are synchronous).
- `randomUUID()` provides 122 bits of entropy — not brute-forceable in a 5-min TTL.
- Memory-only by design; dashboard restart invalidates all outstanding tokens. Correct trade-off for a short-TTL confirmation token.

**Wiring (`handleProjectsAdd`)**
- Path validation runs at line 3697, **before** `safeJsonObj(configPath)` — rejected requests leave zero config mutations. ✓
- Auto-detected `name` (from `package.json` / directory basename) also flows through `validateProjectName` at line 3770. A hostile `package.json#name` can't bypass the allowlist.

**Tests**
- 16 new tests, all passing. Covers the full acceptance-criteria matrix: valid names, too-long, traversal, shell metachars, `.git` dir vs file, non-existent path, non-repo without confirmation, non-repo with wrong token, non-repo with allowNonRepo but no token, non-repo with valid token, and the \"ignore allowNonRepo when path IS a repo\" defense-in-depth case.
- Full suite: **2462 passed / 1 failed / 3 skipped**. The 1 failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`) exists on master too — verified via `git stash`. Net delta: +16 passing, 0 new failures.

## Non-blocking observations (deferred to human judgment)

1. **Unbounded token map.** `_projectConfirmTokens` has no size cap. A client with dashboard access (currently unauthenticated — SEC-01 is in-flight by Rebecca) can mint arbitrary tokens. Each entry is ~50 bytes, purged on next sweep at 5 min, so bounded at `rps * 300s * 50B`. Low severity until dashboard auth lands; worth a cap (e.g. `Map` size > 10k → reject new mints) as a follow-up. **Not blocking** — dashboard auth will gate this endpoint anyway.

2. **Scoped package names with dots now rejected.** Auto-detection strips `@scope/` prefix (`dashboard.js:3750`), so a `package.json` name like `@minions/my.cool.pkg` becomes `my.cool.pkg` — which the new allowlist rejects (dot not in `/^[a-zA-Z0-9_\-]+$/`). Users would need to pass an explicit `name:` override. This is the correct security stance (dots enable path-component abuse), but it's a user-visible behavior change worth calling out. **Not blocking** — override path exists.

3. **Case-insensitive duplicate-project dedup on Windows.** The existing `path.resolve(p.localPath) === target` check in `handleProjectsAdd` remains case-sensitive, so `C:\Foo` and `C:\foo` could both be added on Windows. Pre-existing — not introduced by this PR.

## Deferred to human

None — I don't see any contested calls. Three observations above are straightforward non-blockers.

Review by Minions (Ralph — Engineer)

PR #1299, opened 2026-04-18
fix: auto-fetch PR title on link-pr (closes #1283)

### LINKED ISSUE #1283: link-pr (CC): title should be auto-fetched from remote, not left as user-supplied placeholder
## Bug

When a PR is linked manually via the Command Center using the `link-pr` action, the stored title is whatever the user typed (or left blank) — the real PR title from ADO/GitHub is never fetched.

## Expected behavior

After the PR URL is resolved, the engine should call the remote API (ADO `/_apis/git/repositories/{repo}/pullRequests/{id}` or equivalent GitHub endpoint) and use the returned `title` field, overriding whatever placeholder was passed in.

## Steps to reproduce

1. Open Command Center
2. Use `link-pr` action with a valid ADO PR URL
3. Observe that the linked PR title shown in the dashboard is not the actual PR title from the repo

## Suggested fix

In the `link-pr` handler, after parsing the PR URL, make an API call to fetch PR metadata and set `title` from the response before writing to state.

yemi33:
Closes yemi33/minions#1283

## What was broken

The `/api/pull-requests/link` handler does async enrichment after the initial write — it calls `gh api` (GitHub) or `ado.fetchAdoPrMetadata` (ADO) and merges the response back into `pull-requests.json`. But the title-update was gated on the user-supplied title being empty:

```js
if (!title && prData.title) pr.title = prData.title.slice(0, 120);
```

Any non-empty placeholder — including CC-supplied values — blocked the real remote title from ever being applied. The form label `"Short description (optional — auto-detected from URL)"` confirms the user-typed title was only ever meant as a provisional placeholder until remote enrichment completed.

## The fix

Drop the `!title &&` guard. One-line change in `dashboard.js:4862`:

```js
// Remote title always wins — any user-supplied title is a placeholder (closes #1283)
if (prData.title) pr.title = prData.title.slice(0, 120);
```

Other enrichment fields (`description`, `branch`, `author`) already had the correct semantics — only `title` had this gating bug.

## Why keep the async pattern

Making the fetch synchronous before the initial write would block the HTTP response on a remote API call (network RTT + ADO token acquisition). The existing pattern writes a placeholder, responds fast, then updates — which is fine once the title-override bug is gone.

## Fallback behavior

If `fetchAdoPrMetadata` returns null (ADO token unavailable) or `gh api` throws, `prData` is falsy and the enrichment IIFE returns early — the user-supplied title (or default `"PR #N (polling...)"`) is preserved.

## Test plan

- [x] Wrote failing regression test first (TDD): `PR link enrichment always overrides title from remote (closes #1283)`. It extracts the link handler between route boundaries, asserts the remote-metadata fetch is still called, and asserts the `if (!title && prData.title)` gating pattern is gone.
- [x] Confirmed test fails against pre-fix source.
- [x] Confirmed test passes after the one-line fix.
- [x] Ran full unit suite: 2448 pass, 1 pre-existing failure ("Metrics JSON has valid structure: dallas missing tasksCompleted") unrelated to this change (verified via `git stash` round-trip).

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -4859,7 +4859,8 @@ What would you like to discuss or change? When you're happy, say "approve" and I
           mutateJsonFileLocked(prPath, (prs) => {
             const pr = prs.find(p => p.id === prId);
             if (!pr) return prs;
-            if (!title && prData.title) pr.title = prData.title.slice(0, 120);
+            // Remote title always wins — any user-supplied title is a placeholder (closes #1283)
+            if (prData.title) pr.title = prData.title.slice(0, 120);
             if (prData.description) pr.description = prData.description.slice(0, 500);
             if (!pr.branch && prData.branch) pr.branch = prData.branch;
             if (pr.agent === 'human' && prData.author) pr.agent = prData.author;
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -9011,6 +9011,33 @@ async function testRecentFeatures() {
       'Should support autoObserve (dispatched) vs context-only (polled but no dispatch)');
   });
 
+  await test('PR link enrichment always overrides title from remote (closes #1283)', () => {
+    // Find the link handler — starts at the route path and runs to '{ method:' for next route
+    const start = dashSrc.indexOf("/api/pull-requests/link'");
+    assert.ok(start > 0, 'link handler must exist');
+    // Scan forward until the next route declaration
+    const next = dashSrc.indexOf("{ method: 'POST', path: '/api/pull-requests/delete'", start);
+    assert.ok(next > start, 'delete handler must come after link handler');
+    const linkHandler = dashSrc.slice(start, next);
+
+    // Enrichment must call fetchAdoPrMetadata or gh api to get remote metadata
+    assert.ok(
+      linkHandler.includes('fetchAdoPrMetadata') || linkHandler.includes('gh api'),
+      'link handler must call remote API to fetch PR metadata'
+    );
+    // Title override must NOT be gated on user-supplied title being empty — real title always wins
+    assert.ok(
+      !/if\s*\(\s*!title\s*&&\s*prData\.title\s*\)/.test(linkHandler),
+      'link handler must NOT gate title override on !title — remote title must always win over user placeholder (closes #1283)'
+    );
+    // Title override must still exist, just unconditional on prData.title presence
+    assert.ok(
+      /prData\.title\s*\)\s*pr\.title\s*=/.test(linkHandler)
+        || /pr\.title\s*=\s*prData\.title/.test(linkHandler),
+      'link handler must still assign pr.title from prData.title when remote returns a title'
+    );
+  });
+
   // Plan creation from dashboard
   await test('POST /api/plans/create endpoint exists', () => {
     assert.ok(dashSrc.includes('/api/plans/create'),


yemi33:
## Implementation notes

### Root cause (one line)

`dashboard.js:4862` gated the async-enrichment title update on `!title`, so any non-empty user-supplied title — including the `"PR #N (polling...)"` placeholder and whatever CC typed — permanently blocked the real remote title.

### Why not make the fetch synchronous

The issue's suggested fix says "fetch metadata and set title before writing to state." I kept the existing async pattern because:

1. **Fast HTTP response.** The link-pr response returns in <100ms; remote fetch is 500ms-3s (ADO token acquisition alone can take 1s+).
2. **Resilient to remote flakes.** If the remote API is down or the ADO token is missing, the link still succeeds — the placeholder title stays until a later poll recovers.
3. **Minimal blast radius.** One-line diff vs. a refactor of the control flow. Other fields in the same merge block (`description`, `branch`, `author`) already relied on the async pattern with correct semantics — only `title` had the bug.

The original suggestion would also work but carries more risk; happy to refactor if a reviewer prefers.

### Regression test design

Source-inspection test at `test/unit.test.js:9014`. It:
1. Slices the link handler between `/api/pull-requests/link` and `/api/pull-requests/delete` route declarations.
2. Asserts remote fetch call is still present (`fetchAdoPrMetadata` or `gh api`).
3. Asserts the `if (!title && prData.title)` pattern is **gone**.
4. Asserts title assignment from `prData.title` is still present (unconditionally).

Brittle to route reordering; documented the caveat in the findings file.

yemi33 (commented):
VERDICT: APPROVE

**Self-authorship disclosure:** This PR was authored by a minion (shared yemi33 identity). Deferring any contested judgment to human.

### Change
Single-line fix in `dashboard.js` link-pr enrichment (dashboard.js:4862): removes the `!title &&` guard so remote PR title always overrides the user-supplied placeholder. Matches the issue — user passes a stub title to `link-pr`, then remote fetch should replace it.

### Strengths
- Minimal, surgical fix with a clear inline comment referencing #1283.
- TDD: test in `test/unit.test.js` asserts the `!title &&` gate is gone *and* that the assignment still exists, preventing a future regression in either direction.
- Description cap (`.slice(0, 500)`) and other fields untouched — scope discipline.

### Non-blocking observations
- The test relies on source-string matching (same pattern used elsewhere in `testRecentFeatures`), so it's consistent with project conventions but won't catch a semantic regression where remote fetch silently fails and leaves `prData.title` undefined. The existing `if (prData.title)` guard handles that safely, so not a concern.
- Consider symmetrically updating the `description` handling at some point if users also pass placeholder descriptions — out of scope here.

Review by Minions (Rebecca — Architect)

yemi33:
## No-op — self-authored comment, not a review request

The flagged comment (`id=4273242449`, author `yemi33`) is an **implementation-notes self-disclosure** from the original author agent, not a reviewer's change request. All agents on this project share the `yemi33` GitHub identity, so self-authored implementation notes can look like review findings to the engine's fix-cycle dispatcher.

Reading the comment content:
- It explains *why* the async pattern was kept ("Fast HTTP response", "Resilient to remote flakes", "Minimal blast radius") — justification, not request.
- It offers: "happy to refactor if a reviewer prefers" — conditional offer, no instruction.
- It documents the regression-test design and its brittleness — disclosure, not ask.

No actionable change was requested. PR #1299 has already been approved (VERDICT: APPROVE on the review thread). Making code changes here would be churn.

### Why I'm not refactoring to synchronous fetch

Even if this were a real ask, the author's rationale stands:
- Link-pr handler returns in <100ms; remote ADO fetch is 500ms–3s (token acquisition alone can exceed 1s on cold cache).
- Async pattern survives remote flakes — placeholder title persists until next poll recovers.
- Other fields in the same merge block (`description`, `branch`, `author`) already rely on the async pattern correctly; only `title` had the `!title &&` gate bug, which this PR removed.

No files changed, no tests run, no push.

PR #1300, opened 2026-04-18
Implement: SEC-13: Prototype pollution guard in readBody

yemi33:
## Summary

Implements PRD item **P-e5b2a7d8-b** — a belt-and-braces guard that rejects HTTP POST bodies containing `__proto__`, `constructor`, or `prototype` keys before the parsed body is handed to any handler.

- New pure helper `hasDangerousKey(obj)` in `engine/shared.js`. Arrays are transparent (they don't consume a depth level); object nesting caps at 1.
- Wired into `dashboard.js` `readBody`: on trigger, rejects with `err.statusCode = 400` and the exact message from the spec.
- Updated `catch (e) { jsonReply(res, 500, ...) }` patterns to honour `e.statusCode || 500`, including the SSE stream handler's pre-`writeHead` error path — so the guard consistently surfaces 400 regardless of which handler catches.

### Why this matters

`JSON.parse` itself is safe — it installs `__proto__` as a plain own data property and does not mutate the prototype chain. The risk is *downstream* code that shallow-merges the payload into a target via `Object.assign`, spread, or deep-merge utilities. Any such site can elevate the payload into a prototype write. Rather than audit every call site, we reject poisoned bodies at the single entrypoint.

### Spec compliance (TDD)

Every row of the unit-test table from the PRD is covered in `test/unit.test.js`:

| Input | Expected | Test |
|---|---|---|
| `{__proto__: {}}` | true | flags top-level __proto__ |
| `{constructor: 1}` | true | flags top-level constructor |
| `{prototype: null}` | true | flags top-level prototype |
| `{foo: {__proto__: 1}}` | true | flags one-level-deep __proto__ |
| `{foo: {bar: {__proto__: 1}}}` | false | does NOT recurse beyond one level by design |
| `{}` | false | empty object |
| `null` / `undefined` / primitives | false | non-objects |
| arrays with dangerous elements | true | transparent iteration |
| arrays of primitives | false | safe |

Plus: input-non-mutation assertion, arrays-of-objects tests, clean-body pass-through.

Integration tests in `test/minions-tests.js`:
- `POST /api/command-center` with `{"message":"hi","__proto__":{"polluted":true}}` → **400** with the forbidden-key error.
- `POST /api/command-center` with `{"message":"hi","constructor":{...}}` → **400**.
- Clean body POST (against a fast endpoint that shares `readBody`) → **200** (guard doesn't over-trigger).
- Asserts `Object.prototype` was NOT mutated as a side effect.

## Test plan

- [x] `npm test` — 2463/2464 pass. The single failure is pre-existing (`Metrics JSON has valid structure` reads live `metrics.json` and is state-dependent; unrelated to this PR).
- [x] 17 new `hasDangerousKey` unit tests all pass.
- [x] Behavioural end-to-end check via a mock-`req` harness confirmed the guard rejects with `statusCode=400`, the clean-body path resolves normally, and `Object.prototype` stays unpolluted.
- [ ] Integration tests (`test/minions-tests.js`) will pass once the dashboard is restarted with this branch's code; they currently red against the running master dashboard by design.

## Files changed

- `engine/shared.js` — `hasDangerousKey` helper + export
- `dashboard.js` — `readBody` guard; `e.statusCode`-aware catch blocks; SSE handler pre-headers error path
- `test/unit.test.js` — `testHasDangerousKey` (17 cases) + registration
- `test/minions-tests.js` — 3 integration cases in `testApiEndpoints`

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -1287,7 +1287,20 @@ function readBody(req) {
       reject(new Error('Request body timeout after 30s'));
     }, 30000);
     req.on('data', chunk => { body += chunk; if (body.length > 1e6) { clearTimeout(timeout); reject(new Error('Too large')); } });
-    req.on('end', () => { clearTimeout(timeout); try { resolve(JSON.parse(body)); } catch(e) { reject(e); } });
+    req.on('end', () => {
+      clearTimeout(timeout);
+      let parsed;
+      try { parsed = JSON.parse(body); } catch(e) { reject(e); return; }
+      // Belt-and-braces: reject payloads containing prototype-pollution attack keys
+      // before they reach any downstream Object.assign / spread / deep-merge.
+      if (shared.hasDangerousKey(parsed)) {
+        const err = new Error('Request body contains forbidden key (__proto__, constructor, or prototype)');
+        err.statusCode = 400; // honoured by handler catch blocks so response is 400 regardless of handler
+        reject(err);
+        return;
+      }
+      resolve(parsed);
+    });
     req.on('error', (e) => { clearTimeout(timeout); reject(e); });
   });
 }
@@ -1514,7 +1527,7 @@ const server = http.createServer(async (req, res) => {
         }
       }
       return jsonReply(res, 200, { ok: true, message: 'Completion check ran but no verify task was needed' });
-    } catch (e) { return jsonReply(res, 500, { error: e.message }); }
+    } catch (e) { return jsonReply(res, e.statusCode || 500, { error: e.message }); }
   }
 
   async function handleWorkItemsRetry(req, res) {
@@ -1842,7 +1855,7 @@ const server = http.createServer(async (req, res) => {
         if (content) { try { allArchived.push(...JSON.parse(content).map(i => ({ ...i, _source: project.name }))); } catch {} }
       }
       return jsonReply(res, 200, allArchived);
-    } catch (e) { console.error('Archive fetch error:', e.message); return jsonReply(res, 500, { error: e.message }); }
+    } catch (e) { console.error('Archive fetch error:', e.message); return jsonReply(res, e.statusCode || 500, { error: e.message }); }
   }
 
   async function handleWorkItemsReopen(req, res) {
@@ -3467,7 +3480,7 @@ What would you like to discuss or change? When you're happy, say "approve" and I
       }
       return jsonReply(res, 200, { ok: true, answer: answer + '\n\n(Read-only — changes not saved)', edited: false, actions });
       } finally { _docAbort = null; docChatInFlight.delete(docKey); }
-    } catch (e) { return jsonReply(res, 500, { error: e.message }); }
+    } catch (e) { return jsonReply(res, e.statusCode || 500, { error: e.message }); }
   }
 
   async function handleInboxPersist(req, res) {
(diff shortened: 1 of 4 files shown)


yemi33 (commented):
VERDICT: APPROVE

Clean, focused implementation of SEC-13. Reviewed against CLAUDE.md conventions and the PR scope.

## What's right

**engine/shared.js — `hasDangerousKey`**
- Correct handling of `JSON.parse`'s `__proto__` quirk: it becomes an own enumerable data property, so `Object.keys`/`Object.values` find it (verified by the test using `JSON.parse('{"__proto__":{"x":1}}')`).
- Arrays transparent (don't consume a depth level) — correct, prevents `[{__proto__:...}]` bypass.
- Depth limit of 1 is explicitly documented as intentional with the DoS rationale. Given `readBody` already caps at 1e6 bytes and the primary risk is `Object.assign(target, payload)` / spread (shallow), depth-1 is the right trade-off. Belt-and-braces, not a full sanitizer — the doc comment says so plainly.
- No mutation of input, null/undefined/primitive short-circuits, exported alongside other sanitizers (`sanitizePath`, `validatePid`) — fits the module's shape.

**dashboard.js — readBody guard + statusCode propagation**
- Guard runs after `JSON.parse` succeeds and rejects with `err.statusCode = 400` so every handler's catch returns 400 instead of 500. 12 sites updated to honour `e.statusCode || 500` — consistent, no missed handlers in the diff.
- SSE path (`/api/command-center/stream`) correctly checks `!res.headersSent` and falls back to a plain JSON 400 when the guard trips before SSE headers go out. Good catch — without this, the client would have received an SSE `error` event with a misleading 200 status.

**Tests — TDD-compliant**
- Unit tests (16 cases) cover: top-level + depth-1 for all three keys, depth-2 deliberately NOT flagged (locks in the documented contract), arrays transparent, primitives/null/empty, and an explicit no-mutation assertion.
- Integration tests exercise the full HTTP path, and critically verify `({}).polluted === undefined` after the attack to prove `Object.prototype` was not mutated as a side effect of the test setup itself.
- Clean-body integration test uses a real endpoint (`/api/work-items`) to prove the guard doesn't over-reject, with teardown. No test pollution.

## Non-blocking observations

1. **Depth-1 is documented but worth a KB entry.** Future agents adding deep-merge utilities (not just `Object.assign`) need to know this guard does NOT protect them. The header comment says so; worth reinforcing in `knowledge/conventions/` during consolidation.
2. **Handlers that bypass `readBody`** (e.g. SSE streaming of uploaded files, if any are added later) won't get the guard. Current diff routes all JSON POSTs through `readBody`, so fine today.

No blockers. Ship it.

yemi33:
## Merge Conflicts Resolved

Resolved two conflicts between `work/P-e5b2a7d8-b` and `master`:

### 1. `engine/shared.js` — module exports list
Both sides added exports:
- Branch: `hasDangerousKey`
- Master: `validateProjectName`, `validateProjectPath`

Resolution: kept all three exports (all are used by their respective callers — verified the functions are actually defined in the file).

### 2. `test/unit.test.js` — new test functions + runner invocations
Same additive pattern:
- Branch: `testHasDangerousKey` (17 assertions covering SEC-13 prototype pollution guard)
- Master: `testValidateProjectName` + `testValidateProjectPath` (SEC-04/SEC-05)

Resolution: kept all three test functions and added all three invocations to `main()`.

### Verification
- `node -c` passes on both files
- Full test suite: **2520 passed, 1 failed, 3 skipped**
- All three merged test functions pass individually:
  - 17/17 `hasDangerousKey` assertions pass
  - All `validateProjectName` / `validateProjectPath` assertions pass
- The 1 pre-existing failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`) is a live `engine/metrics.json` data issue unrelated to this merge — identical test exists on master.

Commit: `164d974c` — `merge: resolve conflicts with master for SEC-13 PR #1300`

PR #1316, opened 2026-04-18
fix: Pinned context optimistic save (closes #1295)

### LINKED ISSUE #1295: Pinned context: manually writing entry should optimistically save and appear in list
## Summary

When a user manually types/writes a pinned context entry, it should optimistically save and immediately appear in the pinned context list — without requiring a page refresh or explicit submit confirmation.

## Expected behavior
- User types a pinned context entry
- Entry appears in the list immediately (optimistic update)
- Save is confirmed in the background

## Current behavior
- Manually written pinned context does not show in the list until after a save/reload cycle

## Notes
UX parity with other list-based inputs in the dashboard.

yemi33:
Closes #1295.

## Summary

When a user pins a note, the entry now appears in the list **immediately** instead of waiting up to ~64 s for the next status refresh to bubble back the new state.

## Why it was slow

Two independent gaps compounded:

1. **Client:** `submitPinnedNote` awaited `fetch('/api/pinned')` before any UI change. The list only updated on the next 4 s `refresh()` tick.
2. **Server:** `pinned.md` is rebuilt into `_slowState` (60 s TTL). Even with a fast `refresh()`, the newly-persisted entry was invisible for up to a minute because `invalidateStatusCache()` only cleared fast state.

### `dashboard/js/render-pinned.js`
- `submitPinnedNote` now snapshots `window._pinnedEntries`, appends the new entry, calls `renderPinned()` synchronously, and only then fires the POST.
- On non-2xx or network failure: revert to the snapshot, re-render, show the error toast, and re-open the modal so the user doesn't lose their input.

### `dashboard.js`
- `invalidateStatusCache(opts)` accepts an opt-in `{ includeSlow: true }` flag that clears `_slowState` alongside `_fastState`. Default callers (dispatch mutations, work-item updates, etc.) are unchanged — slow state continues to use its own 60 s TTL.
- `POST /api/pinned` and `POST /api/pinned/remove` both opt in, so a subsequent `/api/status` rebuild surfaces the new state on the next tick rather than waiting for TTL expiry.

## Tests

Four new tests in `test/unit.test.js` assert:
- `renderPinned` fires **before** `fetch` (optimistic ordering).
- A snapshot is captured and `renderPinned` is called a second time on the failure path (revert).
- Both `POST /api/pinned` and `POST /api/pinned/remove` call `invalidateStatusCache({ includeSlow: true })`.
- `invalidateStatusCache` still gates `_slowState = null` behind the `includeSlow` branch (existing tests updated to reflect the new signature).

All 2469 unit tests pass.

## Test plan
- [ ] Pin a note in the dashboard — verify it appears instantly in the list without any visible refresh pause.
- [ ] Unpin a note — verify removal is instant and that a fresh `/api/status` fetch doesn't resurrect it for up to 60 s.
- [ ] Force a server 500 on `POST /api/pinned` — verify the optimistically-added entry disappears, the modal re-opens with the typed values, and the failure toast shows.

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -381,10 +381,15 @@ function _mtimesChanged(prev, curr) {
   return false;
 }
 
-function invalidateStatusCache() {
+function invalidateStatusCache(opts) {
   _fastState = null;
   _fastStateTs = 0;
-  // Slow state continues on its own TTL — not invalidated by mutations
+  // Slow state continues on its own TTL by default — mutations of slow-state data
+  // (pinned.md, schedules, etc.) must opt in via { includeSlow: true } for immediate visibility.
+  if (opts && opts.includeSlow) {
+    _slowState = null;
+    _slowStateTs = 0;
+  }
   _statusCache = null;
   _statusCacheJson = null;
   _statusCacheGzip = null;
@@ -4726,7 +4731,8 @@ What would you like to discuss or change? When you're happy, say "approve" and I
       const levelTag = level === 'critical' ? '🔴 ' : level === 'warning' ? '🟡 ' : '';
       const entry = '\n\n### ' + levelTag + title + '\n\n' + content + '\n\n*Pinned by human on ' + new Date().toISOString().slice(0, 10) + '*';
       safeWrite(pinnedPath, (existing || '# Pinned Context\n\nCritical notes visible to all agents.') + entry);
-      invalidateStatusCache();
+      // pinned.md is in slow-state cache — opt-in invalidation so the new entry is visible immediately (closes #1295)
+      invalidateStatusCache({ includeSlow: true });
       return jsonReply(res, 200, { ok: true });
     }},
     { method: 'POST', path: '/api/pinned/remove', desc: 'Remove a pinned note by title', params: 'title', handler: async (req, res) => {
@@ -4739,7 +4745,8 @@ What would you like to discuss or change? When you're happy, say "approve" and I
       const regex = new RegExp('\\n\\n###\\s*(?:🔴\\s*|🟡\\s*)?' + title.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '\\n[\\s\\S]*?(?=\\n\\n###|$)', 'i');
       content = content.replace(regex, '');
       safeWrite(pinnedPath, content);
-      invalidateStatusCache();
+      // pinned.md is in slow-state cache — opt-in invalidation so the unpin is visible immediately
+      invalidateStatusCache({ includeSlow: true });
       return jsonReply(res, 200, { ok: true });
     }},
 
--- a/dashboard/js/render-pinned.js
+++ b/dashboard/js/render-pinned.js
@@ -42,11 +42,37 @@ async function submitPinnedNote(e) {
   const level = document.getElementById('pin-level').value;
   if (!title || !content) { if (btn) { btn.disabled = false; btn.textContent = 'Pin Note'; } alert('Title and content required'); return; }
   try { closeModal(); } catch { /* may not be open */ }
+
+  // Optimistic render: append the new entry to the pinned list and re-render immediately
+  // so it appears without waiting for the POST round-trip or the next status refresh (closes #1295).
+  // Snapshot prevEntries so we can revert on failure.
+  const prevEntries = Array.isArray(window._pinnedEntries) ? window._pinnedEntries.slice() : [];
+  const newEntry = { title, content, level: level || 'info' };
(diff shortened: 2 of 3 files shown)


yemi33:
## Implementation notes

**Root cause was dual-layered.** The UI never attempted optimistic rendering, and even if it had done a post-fetch refresh, the server's two-tier status cache (`_slowState`, 60 s TTL) would have served stale pinned entries for up to 60 s because `invalidateStatusCache()` only cleared `_fastState`.

**Design choice — opt-in slow-state invalidation.** I considered two alternatives:

1. Always clear `_slowState` from `invalidateStatusCache()`. Rejected — every dispatch tick, work-item update, and metric write would bust the slow-state cache, defeating the whole point of the two-tier split (skills, PRDs, projects, version, etc. change rarely).
2. Expose `invalidateSlowState()` as a separate helper. Workable, but forces every caller to remember two functions. An `opts` bag keeps the surface tight and self-documenting (`invalidateStatusCache({ includeSlow: true })` reads naturally).

The opt-in flag is the right seam: slow-state callers are a closed set (pinned, skills, schedules, etc.), and each decides explicitly whether a mutation warrants slow-state bust. Default callers remain unchanged and the TTL contract is preserved.

**Shape compatibility.** `parsePinnedEntries()` emits `{ level, title, content }`; the client's optimistic entry is `{ title, content, level: level || 'info' }`. Key order differs but field names and types match, so `_changed('pinned', …)` JSON-hash diffing will correctly detect the subsequent server reconciliation and re-render once.

**Revert semantics.** On failure (non-2xx or network error) the client restores the exact pre-submit snapshot of `window._pinnedEntries` and re-renders. Input isn't lost — `openPinNoteModal()` is re-opened so the user can retry or edit. This matches the pattern already used for optimistic unpin (`removePinnedNote` pre-removes the DOM card and calls `refresh()` on failure).

**Test coverage rationale.** Tests use source-code string matching because the UI logic is JS loaded in a browser-like DOM that's not set up in the unit harness. Behavioural tests (JSDOM round-trip) would be ideal but out of scope for this one-line bug. The asserts lock in the ordering (`renderPinned` before `fetch`), the snapshot capture, and both server-side opt-in sites — enough to catch regressions.

yemi33 (commented):
VERDICT: APPROVE

Clean, focused fix for #1295.

**Server (dashboard.js):**
- `invalidateStatusCache(opts)` extended with opt-in `includeSlow` — good API shape; default TTL behaviour preserved for other mutations (no performance regression).
- Both `/api/pinned` (add) and `/api/pinned/remove` pass `{ includeSlow: true }` — correct, since pinned.md lives in slow state.

**Client (render-pinned.js):**
- Optimistic render happens before `fetch`, with `prevEntries` snapshot and full revert on both non-ok response and network error.
- `renderPinned` calls wrapped in try/catch — safe against missing DOM.
- Toast ordering follows the project convention (success before API call, error overwrites on failure).

**Tests:** Four well-scoped assertions cover: optimistic render-before-fetch ordering, revert-on-failure snapshot + re-render, server-side `includeSlow: true` on both pinned routes, and the gated `_slowState = null` branch. Existing cache-tier tests updated to tolerate the new signature.

Minor (non-blocking): the `submitPinnedNote` regex in tests matches up to `window.MinionsPinned` or next function — fine today, but future refactors should keep that anchor present. Not worth a change.

Review by Minions (Ralph — Engineer)

yemi33:
## No-op — flagged "review comment" is self-authored implementation notes

The comment the engine flagged as review feedback (https://github.com/yemi33/minions/pull/1316#issuecomment-4273265625) was posted by **me (Rebecca)** at 08:46 UTC, ~7 minutes after I opened this PR at 08:39 UTC. It's the implementation-notes comment I wrote per the playbook.

Evidence:
- Comment author: `yemi33` — the shared `gh` CLI identity every agent in this team uses (see `pinned.md` → "GitHub Self-Approval is Blocked").
- `gh pr view 1316 --json comments` returns `viewerDidAuthor: true` for comment id `IC_kwDORlkta87-tNvZ`.
- Content is verbatim the "Implementation notes" block I posted after `gh pr create` — all five sections (root cause, design choice, shape compatibility, revert semantics, test coverage rationale) are my own rationale, not reviewer feedback.
- There is no imperative, no question, no request for change. It's a defense of the chosen design, already self-consistent with the PR diff.

No code changes are appropriate here: there is nothing to act on. I am deliberately leaving the PR untouched to avoid (a) double-commenting on my own design rationale, (b) introducing churn, and (c) burning review cycles on a self-handoff artifact.

If a human reviewer wants changes, they can post fresh feedback and the engine will re-dispatch.

PR #1327, opened 2026-04-18
Implement P-e5b2a7d8-a: Header gate — Origin allowlist, Content-Type enforcement, security response headers

yemi33:
## Summary

Closes browser-side CSRF / DNS-rebind attack paths on the dashboard HTTP server (`dashboard.js`).

- **Origin allowlist** on all mutating methods (POST/PUT/PATCH/DELETE). Cross-origin requests return 403 JSON. Referer fallback when Origin is absent. Legacy curl/tooling with neither header still passes (preserves local automation).
- **Content-Type enforcement**: mutating requests carrying a body must set `application/json`. Tightened past the spec's literal "absent or text/plain" rule to also block `x-www-form-urlencoded` / `multipart/form-data` — the full simple-request CSRF loophole. Bodyless DELETE (e.g. `/api/cc-sessions/:tabId`) is exempt.
- **Security response headers** set once at dispatcher entry — `Content-Security-Policy` (strict `script-src 'self'`), `X-Frame-Options: DENY`, `X-Content-Type-Options: nosniff`, `Referrer-Policy: same-origin`. Dashboard HTML entry-point overrides CSP with inline-permissive values; all `/api/*` responses inherit the strict baseline (verified by integration test).
- **SSE Origin gate** in `handleAgentLiveStream` (GET) and `handleCommandCenterStream` (POST SSE) — validates Origin BEFORE `res.writeHead(200, text/event-stream)` so rejections are a normal 403 status line, not a dropped stream upgrade.
- `Access-Control-Allow-Origin: '*'` removed from `jsonReply` — set only on GET/HEAD responses by the dispatcher prelude. OPTIONS preflight echoes the validated Origin + `Vary: Origin` instead of `*`.
- **Pure helpers extracted** to `engine/shared.js`: `isAllowedOrigin(origin)` (port-agnostic WHATWG parse, IPv6-bracket aware, rejects literal `'null'`, subdomain-spoofs, non-http schemes) and `buildSecurityHeaders()` (fresh object per call).

## Test plan

- [x] Unit tests — `isAllowedOrigin` truth table (allowed, disallowed, null, IPv6 `[::1]`, malformed, subdomain spoofs, Referer-style URLs) + `buildSecurityHeaders` invariants (keys, strict script-src, fresh-object). `npm test` → 2464 passed / 1 pre-existing failure (unrelated metrics test).
- [x] Integration tests — 11 scenarios in `test/minions-tests.js`:
  - `POST /api/command-center` with `Origin: https://evil.example` → 403
  - `POST /api/work-items` with valid localhost Origin → 200
  - `POST /api/command-center` with `Content-Type: text/plain` → 415
  - `POST` with missing Content-Type (body present) → 415
  - `POST` with `application/json` → passes Origin + CT gates
  - `POST /api/work-items` with evil Origin → 403
  - `GET /api/agent/:id/live-stream` with cross-origin → 403 (JSON, not SSE)
  - `GET /api/status` includes all 4 security headers
  - `POST` with no Origin, no Referer → allowed (legacy curl)
  - `POST` with disallowed Referer (no Origin) → 403
- [x] Manual smoke against side-car dashboard on :7332 confirmed every gate. `BASE` in `test/minions-tests.js` now reads `MINIONS_TEST_BASE` env var so integration coverage can target a non-default port.

## Run instructions

```bash
# Unit tests
npm test

# Integration tests (needs dashboard running on 7331)
minions start
npm run test:all
```

### DIFF
--- a/dashboard.js
+++ b/dashboard.js
@@ -1305,7 +1305,10 @@ function checkRateLimit(key, maxPerMinute) {
 
 function jsonReply(res, code, data, req) {
   res.setHeader('Content-Type', 'application/json');
-  res.setHeader('Access-Control-Allow-Origin', '*');
+  // Access-Control-Allow-Origin is set ONCE by the server dispatcher prelude:
+  // `*` for GET/HEAD (read-only), never for mutating responses (Origin gate
+  // already blocked cross-origin POSTs). Setting it here would reopen the
+  // cross-origin write path.
   res.statusCode = code;
   const json = JSON.stringify(data);
   const ae = req && req.headers && req.headers['accept-encoding'] || '';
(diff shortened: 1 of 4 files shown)


yemi33 (commented):
VERDICT: APPROVE

Solid defense-in-depth header/origin gate. Reviewed the full diff end-to-end.

**Self-authorship disclosure:** All agents share `yemi33` gh credentials, so I may be a teammate of the author. Posting as comment (not vote) per pinned policy.

**What's right**
- `isAllowedOrigin` parses via WHATWG URL (robust against spoofing like `localhost.evil.com`, `localhost.nip.io`), explicitly rejects literal `"null"`, non-http schemes, and non-string inputs. Tests cover all of these (test/unit.test.js:374-500).
- Port-agnostic allowlist is correct (browser tabs may hit localhost/127.0.0.1/[::1]:any-port).
- Origin gate runs BEFORE `res.writeHead(200, text/event-stream)` on both SSE endpoints (dashboard.js:2355-2366, 4042-4052) — critical, since a post-upgrade 403 is indistinguishable from a dropped stream.
- Content-Type 415 gate closes the CSRF simple-request loophole (text/plain, urlencoded, multipart). Empty-body DELETE exemption is narrow and justified.
- `jsonReply` no longer blanket-sets `ACAO: *`; dispatcher echoes `*` only on GET/HEAD. Mutating responses correctly omit ACAO — same-origin browsers still work, cross-origin is denied at two layers (Origin gate + no ACAO).
- HTML entry-point CSP is relaxed ONLY for that response (inline scripts/styles + onclick handlers); API JSON/SSE inherits strict CSP with no `'unsafe-inline'` on script-src. Test asserts this (test/minions-tests.js:491).
- Legacy no-Origin/no-Referer path preserved for curl/Node tooling, documented with rationale.
- Referer fallback only when Origin absent — correct precedence.

**Minor non-blocking observations (do not fix in this PR)**
1. `isAllowedOrigin` rejects `https://` unconditionally. Future reverse-proxy / Tailscale Funnel scenarios would need this revisited. Documented intent is fine for now.
2. The SSE origin block at dashboard.js:2355 duplicates logic from the top-level dispatcher but only because GETs normally skip the gate. The "belt-and-suspenders" comment at 4042 makes the pattern explicit; worth extracting a `rejectDisallowedSseOrigin(req, res)` helper next pass to keep the two sites in sync.
3. `_originGateReject` closure captures `_rawOrigin`/`_rawReferer` — fine today, but if this file ever gets split the helper should move to shared.js alongside `isAllowedOrigin`.
4. Content-Type check uses `startsWith('application/json')` — correctly accepts `; charset=utf-8`. Good.
5. Integration test #8 uses `/api/agent/dallas/live-stream` — works because Origin check fires before the agent-dir existence check, but if that ordering ever flips the test would false-positive. Consider a sentinel agent ID.

**Verified**
- Lock/concurrency rules: N/A (no shared-JSON mutation in this PR).
- Constants usage: N/A (HTTP status codes, not WI_STATUS territory).
- Cross-platform: pure string/URL parsing, no process/path issues.
- Tests: 11 integration + ~20 unit, covering allow/deny matrix, IPv6, Referer fallback, SSE pre-upgrade rejection, and CSP strictness on API vs HTML.

Ready to merge.

yemi33:
## Merge conflicts resolved

Merged `master` into `work/P-e5b2a7d8-a` (commit `9f94f6c8`).

### Conflict analysis — `engine/shared.js`

Both sides added security helpers at the same location in the file. The additions are **independent and complementary**, so I kept both.

| Side | Addition | Purpose |
|------|----------|---------|
| `HEAD` (this PR) | `isAllowedOrigin`, `buildSecurityHeaders` | Header gate (Origin allowlist + security response headers) |
| `origin/master` | `hasDangerousKey`, `validateProjectName`, `validateProjectPath` | SEC-13 prototype pollution + SEC-04/05 project name/path validation |

Both helper sections were retained in their original positions, and the module export block now unions both export sets (`isAllowedOrigin, buildSecurityHeaders, hasDangerousKey, validateProjectName, validateProjectPath`).

### Verification

- `node -c engine/shared.js` → syntax OK
- `npm test` → **2552 pass / 0 fail / 3 skipped**
- `gh pr view 1327` → `mergeable: MERGEABLE` (was `CONFLICTING`)

No code in this PR changed — only the merge commit. The header gate implementation is unchanged from its last-approved state.

PR #1390, opened 2026-04-19
test(scheduler): add unit tests for scheduler.js

yemi33:
Adds coverage for previously untested functions in engine/scheduler.js.

Part of daily coverage rotation — target: engine/scheduler.js

## What's covered

33 new tests in `test/unit.test.js`, grouped under a new
`testSchedulerAdditionalCoverage()` section registered in `main()` after the
pre-existing scheduler suites.

### `parseCronField`
- single exact value
- step syntax: zero-step → always-false, non-numeric step → always-false
- list syntax: drops NaN entries silently, tolerates whitespace around entries
- wholly unparseable syntax → always-false
- trims surrounding whitespace on a single value

### `parseCronExpr`
- dayOfWeek field gates the match (Monday-only vs Tuesday)
- step in hour field (`*/6`)
- list in dayOfWeek (`1,3,5` = Mon/Wed/Fri)
- collapses multiple spaces between fields
- rejects non-string inputs (number, object, array, boolean)

### `shouldRunNow`
- false when cron is invalid (unparseable, null, missing, too-few fields)
- false when current time doesn't match cron (dayOfWeek mismatch)
- true when matching cron and lastRunAt is null / undefined / unparseable
- true when lastRunAt is in a previous calendar minute
- false when lastRunAt is in the same calendar minute

### `discoverScheduledWork`
- `[]` when `config.schedules` is missing / non-array / empty
- skips schedules missing `id`, `cron`, or `title`
- skips schedules with `enabled: false`
- `enabled: undefined` / `null` default to enabled (strict false check)
- skips schedules whose cron doesn't match now
- work-item shape with sane defaults: `type`/`priority`/`description`/`agent`/`project`
- carries over custom fields when present
- legacy string-format `lastRun` entry is honored for same-minute dedup
- legacy string-format entry is upgraded to object shape on dispatch
- mixed set: only fires matching + enabled + well-formed entries
- preserves existing object-shape fields (`lastResult`, `lastCompletedAt`)
  across the dispatch-time re-write

## Test patterns reused

- Date-stubbing via class extending `origDate`
  ([test/unit.test.js:11477-11490](test/unit.test.js#L11477))
- `snapshotAndClearScheduleRuns` (plus `.backup` sidecar)
  pattern derived from `_snapshotAndClearScheduleRuns`
  ([test/unit.test.js:11531-11543](test/unit.test.js#L11531))
- All schedule IDs prefixed with `W-mo5u52v2aabv-` to avoid collision
  with real engine state

## Test results

```
2537 passed  0 failed  3 skipped
```

All 33 new tests pass. No source changes — tests only.

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -11628,6 +11628,505 @@ async function testSchedulerDispatchTimePersistence() {
   });
 }
 
+// ─── scheduler.js — Additional coverage (W-mo5u52v2aabv) ───────────────────
+
+async function testSchedulerAdditionalCoverage() {
+  console.log('\n── scheduler.js — additional parseCron/shouldRunNow/discoverScheduledWork coverage ──');
+
+  // Snapshot + clear schedule-runs.json (and its .backup sidecar). safeJson auto-
+  // restores from .backup when the main file is missing, so both must be cleared
+  // for a true blank slate inside tests. Returns a restore function that puts
+  // whatever was there back — keeps tests atomic with respect to real engine state.
+  function snapshotAndClearScheduleRuns() {
+    const realPath = scheduler.SCHEDULE_RUNS_PATH;
+    const backupPath = realPath + '.backup';
+    const snapshot = fs.existsSync(realPath) ? fs.readFileSync(realPath) : null;
+    const backupSnap = fs.existsSync(backupPath) ? fs.readFileSync(backupPath) : null;
+    try { fs.unlinkSync(realPath); } catch {}
+    try { fs.unlinkSync(backupPath); } catch {}
+    return function restoreFiles() {
+      if (snapshot) fs.writeFileSync(realPath, snapshot);
+      else { try { fs.unlinkSync(realPath); } catch {} }
+      if (backupSnap) fs.writeFileSync(backupPath, backupSnap);
+      else { try { fs.unlinkSync(backupPath); } catch {} }
+    };
+  }
+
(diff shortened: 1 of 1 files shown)


yemi33 (commented):
VERDICT: APPROVE

30 new behavioral tests for `engine/scheduler.js` covering `parseCronField`, `parseCronExpr`, `shouldRunNow`, and `discoverScheduledWork`. All assertions match the source in `engine/scheduler.js`:

- Edge-case cron syntax (`*/0`, `*/abc`, `@@foo`, whitespace lists) correctly map to the `() => false` fall-through.
- `enabled === false` strict check verified for `undefined`/`null`/missing cases (schema default = enabled).
- Legacy string-shape and object-shape `runs[id]` entries both tested, including the upgrade-on-dispatch path and the preservation of `lastResult`/`lastCompletedAt` across the spread rewrite.
- Calendar-minute dedup (same-minute suppresses, prior-minute fires) exercises the regression fix from W-mo3zu273f8tm.
- Dates (Mar 30 2026 = Mon, Mar 31 = Tue, Apr 1 = Wed, Apr 3 = Fri, Apr 5 = Sun) spot-checked and correct.

Good hygiene: `snapshotAndClearScheduleRuns` clears both `schedule-runs.json` and its `.backup` sidecar (avoids the `safeJson` auto-restore footgun), and `withFixedNow` uses try/finally so `global.Date` is restored on assertion failure. Both helpers are documented inline.

No blocking issues. Nothing to change.

PR #1735, opened 2026-04-20
test(preflight): add unit tests for preflight checks

yemi33:
## Summary

Adds coverage for previously untested behavior in `engine/preflight.js`.

Previously, the two existing preflight suites (`testPreflightModule`, `testPreflightDeep`) leaned heavily on source-string matching — asserting that the source code contained certain strings rather than exercising the actual code paths. This PR adds a new `testPreflightBehavioral()` suite that runs the real code with injected failures and captured I/O:

- **`findClaudeBinary`**: MINIONS_DEBUG-gated logging (both enabled and silenced), null-return path with all search candidates falsified.
- **`runPreflight`**: Git-missing and Claude-CLI-missing failure branches via a `withStubbedExecSync` helper that mutates `child_process.execSync` AND busts the preflight require-cache (preflight destructures `execSync` at module load, so the re-require is mandatory).
- **`checkOrExit`**: actual process.exit(1) behavior under `exitOnFail=true` + failure; no-exit under pass; `exitOnFail=false` returns false without exiting; custom label propagates to printPreflight output.
- **`doctor`**: full behavioral matrix against a tmp MINIONS_HOME — engine.js present/missing, config.json absent/malformed, placeholder-only projects, missing project localPath, empty/populated agents block, Teams integration (client-secret auth, certificate auth, incomplete, disabled), Engine state (running with stale PID, stopped, control.json missing), Playbooks OK/empty, Port 7331 conditional row, summary line format.

29 new tests — all pass. One pre-existing test (`Metrics JSON has valid structure`) still fails because the live `engine/metrics.json` has a malformed `dallas` entry — confirmed pre-existing via `git stash && npm test` on master (2503 passed, 1 failed identical).

Part of daily coverage rotation — target: `engine/preflight.js`.

## Test plan

- [x] `npm test` — 2532 passed, 1 pre-existing failure, 3 skipped
- [x] New behavioral suite (29 tests) all green
- [x] No regressions introduced by the new tests
- [x] Test isolation verified — require.cache and process.env restored in every finally block

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -4855,6 +4855,686 @@ async function testPreflightDeep() {
   });
 }
 
+// ─── engine/preflight.js Behavioral Tests ───────────────────────────────────
+//
+// These tests exercise real code paths (failure injection, stubbed process.exit,
+// fake MINIONS_HOME for doctor) rather than matching on source strings. The
+// two suites above are source-string heavy; these cover the actual branches.
+
+async function testPreflightBehavioral() {
+  console.log('\n── engine/preflight.js (behavioral) ──');
+
+  const PREFLIGHT_ABS = require.resolve(path.join(MINIONS_DIR, 'engine', 'preflight'));
+
+  // Load fresh each test so env/execSync stubs apply.
+  function freshPreflight() {
+    delete require.cache[PREFLIGHT_ABS];
+    return require(PREFLIGHT_ABS);
+  }
+
+  // Capture console.log/error; also stub process.exit so failure paths don't nuke the runner.
+  function withCapture(fn) {
+    const logs = [];
+    const errs = [];
+    const exits = [];
+    const origLog = console.log;
+    const origErr = console.error;
+    const origExit = process.exit;
+    console.log = (...a) => { logs.push(a.map(String).join(' ')); };
+    console.error = (...a) => { errs.push(a.map(String).join(' ')); };
+    process.exit = (code) => { exits.push(code); throw new Error('__EXIT__' + code); };
+    let threw = null;
+    let result;
+    try { result = fn(); }
+    catch (e) { if (!/^__EXIT__/.test(e.message)) threw = e; }
+    finally {
+      console.log = origLog;
+      console.error = origErr;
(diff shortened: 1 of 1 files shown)


yemi33 (commented):
VERDICT: APPROVE

Solid behavioral test suite for engine/preflight.js. Covers the gaps left by the existing source-string-heavy `testPreflightModule`/`testPreflightDeep` suites.

## What's good
- **Real execution paths**: stubs `child_process.execSync` via module mutation + fresh `require.cache` invalidation, so the destructured `{ execSync }` binding at top of `preflight.js` actually picks up the stub. Correct technique.
- **Safe `process.exit` stub**: throws sentinel `__EXIT__<code>` and filters it out of `threw` — prevents the runner from dying on `checkOrExit`.
- **Host-tolerant guards**: `findClaudeBinary`/Claude-CLI assertions allow the "binary actually exists on dev box" case through (`if (result !== null) …`). Good defensive test design — avoids flakes on machines with claude-code globally installed.
- **Structured output parsing**: `parseResultLine` against the ✓/✗/! icons lets the doctor tests assert per-check outcomes cleanly rather than substring-matching the whole blob.
- **Fake MINIONS_HOME fixtures**: `setupDoctorHome` with tunable `config`/`control`/`playbooks` is the right shape for exercising every doctor branch (placeholder project, missing localPath, empty agents, teams auth modes, stale PID, playbook count, summary line).
- **Env var hygiene**: each test saves/restores `MINIONS_DEBUG`, `npm_config_prefix`, `APPDATA` in `finally` blocks — no cross-test pollution.
- **Cache cleanup**: final `delete require.cache[PREFLIGHT_ABS]` ensures downstream test suites get an un-stubbed preflight.

## Minor non-blocking notes
- Stale-PID test uses `pid: 999999` — fine on most hosts, but on a long-uptime Linux box PIDs can reach that range. Low probability, acceptable.
- `withStubbedExecSync` restores `cp.execSync` in finally but also invalidates the require cache — subtle but correct: any cached preflight would still hold the stubbed binding otherwise.

Non-blocking. Ready to merge.

PR #1736, opened 2026-04-20
test(meeting): add unit tests for meeting lifecycle

yemi33:
## Summary

Adds coverage for previously untested functions in `engine/meeting.js`.

Part of daily coverage rotation — target: `engine/meeting.js`.

## What changed

**Refactor — `engine/meeting.js`:**
- `MEETINGS_DIR` now derives from `shared.MINIONS_DIR` instead of a hardcoded `__dirname` join.
- `DISPATCH_PATH` inside `_killMeetingDispatches` likewise routed through `shared.MINIONS_DIR`.

This makes the module honor `MINIONS_TEST_DIR`, so `createTestMinionsDir()` can isolate meeting state cleanly — matching the pattern already used by `shared.js`, `cleanup.js`, `llm.js`, etc.

**Tests — `test/unit.test.js`:**

New block `testMeetingsIsolatedGaps()` with 13 tests filling gaps left by the existing three meeting-test blocks:

- `EMPTY_OUTPUT_PATTERNS` — verifies exported array shape, contents (`(no output)`, `(no findings)`, `(no response)`), and that real findings merely mentioning a placeholder as a substring are **accepted** (exact-match semantics, not regex).
- `advanceMeetingRound` — archived meeting returns `null`; unknown status (e.g. `paused`) returns the meeting unchanged without bumping `roundStartedAt`.
- `addMeetingNote` — multiple notes persist in call order; pre-existing `humanNotes` are preserved.
- `checkMeetingTimeouts` — meeting with no `roundStartedAt` is a no-op even when the configured timeout is 1ms.
- `collectMeetingFindings` — unknown `roundName` is a silent no-op (no findings/debate/conclusion mutation).
- `discoverMeetingWork` — no new work items when every participant has already submitted for the current round.
- `MEETINGS_DIR` — resolves relative to `shared.MINIONS_DIR` under `createTestMinionsDir()` (pins the refactor).
- `deleteMeeting` — removes file AND moves matching active dispatch entries to `completed` with `result: 'error'`, while leaving unrelated dispatch entries untouched.
- `deleteMeeting` on a missing meeting — returns `false` and does not touch `dispatch.json`.

## Test plan

- [x] `npm test` runs from project root: **2517 passed, 0 failed, 3 skipped** (was 2504 passing before the new block).
- [x] All new tests clean up after themselves via `finally` blocks — either `cleanupMeeting(id)` on the real meetings dir, or `restore()` from `createTestMinionsDir()`.
- [x] No test-only agent IDs written to real state (confirmed by existing `Test Isolation Verification` block).
- [x] `engine/meeting.js` refactor is behavior-preserving: `path.join(shared.MINIONS_DIR, 'meetings')` resolves to the identical path as the old `path.join(__dirname, '..', 'meetings')` when `MINIONS_TEST_DIR` is unset.

### DIFF
--- a/engine/meeting.js
+++ b/engine/meeting.js
@@ -16,7 +16,9 @@ const EMPTY_OUTPUT_PATTERNS = ['(no output)', '(no findings)', '(no response)'];
 
 // No lazy require needed — log comes from shared.js, no engine-specific APIs used
 
-const MEETINGS_DIR = path.join(__dirname, '..', 'meetings');
+// Derive from shared.MINIONS_DIR so createTestMinionsDir()/MINIONS_TEST_DIR
+// tests can redirect the meetings directory without patching module internals.
+const MEETINGS_DIR = path.join(shared.MINIONS_DIR, 'meetings');
 
 function truncateMeetingContext(text, maxBytes, label) {
   return shared.truncateTextBytes(text, maxBytes, `\n\n_...${label} truncated — review the meeting transcript if needed._`);
@@ -403,7 +405,7 @@ function addMeetingNote(meetingId, note) {
 
 function _killMeetingDispatches(meetingId) {
   try {
-    const DISPATCH_PATH = path.join(__dirname, '..', 'engine', 'dispatch.json');
+    const DISPATCH_PATH = path.join(shared.MINIONS_DIR, 'engine', 'dispatch.json');
     const dispatch = safeJson(DISPATCH_PATH) || {};
     const toKill = (dispatch.active || []).filter(d => d.meta?.meetingId === meetingId);
     if (toKill.length === 0) return 0;
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -11388,6 +11388,341 @@ async function testMeetingsExtendedBehavioral() {
   });
 }
 
+// ─── Meeting — Isolated Gap Tests (W-mo79lrbnkc71) ──────────────────────────
+// Coverage for previously untested behaviors in engine/meeting.js:
+//   - EMPTY_OUTPUT_PATTERNS export shape + match semantics
+//   - advanceMeetingRound: archived meeting returns null; no-change branch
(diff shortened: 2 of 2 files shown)


yemi33 (commented):
VERDICT: APPROVE

Solid gap-coverage PR. The two production-code changes are minimal and correct:

- `engine/meeting.js:21` — MEETINGS_DIR now routes through `shared.MINIONS_DIR` instead of `__dirname`, enabling MINIONS_TEST_DIR isolation.
- `engine/meeting.js:408` — `_killMeetingDispatches` DISPATCH_PATH same treatment. Resolves to the same absolute path in production (shared.MINIONS_DIR is the repo root), so no behavior change.

Tests (13 new, all passing locally):
- EMPTY_OUTPUT_PATTERNS shape + exact-match semantics (the substring test is a nice guard against someone regex-ifying this later).
- advanceMeetingRound archived + unknown-status branches (verifies the no-change branch does not rewrite roundStartedAt — good touch).
- addMeetingNote ordering + seeded-notes preservation, including transcript entries.
- checkMeetingTimeouts no-op when roundStartedAt is missing.
- collectMeetingFindings silent no-op for unknown roundName (findings/debate/conclusion all untouched).
- discoverMeetingWork returns no new work when every participant has submitted.
- Isolated MINIONS_TEST_DIR test proves the path refactor.
- deleteMeeting + _killMeetingDispatches integration: kills only matching active entries, preserves unrelated + bare-meta entries, marks killed as error with reason + completed_at; and the no-op false-return case when file is absent.

Non-blocking observations:
- The tests use unique TEST-GAP-* IDs and `finally { cleanupReal(id) }` for non-isolated runs — appropriate, though a mid-test crash could still leak a file into the real `meetings/` dir. Low risk given the deterministic IDs.

Full suite: 2516 passed, 1 pre-existing unrelated EPERM failure on Windows (readonly-primary), 3 skipped — not introduced by this PR.

PR #1738, opened 2026-04-21
test(llm): add unit tests for trackEngineUsage and isResumeSessionStillValid

yemi33:
## Summary

Adds 19 new unit tests covering previously untested behavior in `engine/llm.js`. Part of daily coverage rotation — target: `engine/llm.js`.

### `trackEngineUsage(category, usage)`

Covers:
- New-category initialization with all counter fields at 0 + 1
- Accumulation across multiple calls (calls, tokens, cost, cache, cacheCreation)
- Missing usage fields default to 0 (no NaN propagation)
- `durationMs` increments `totalDurationMs` and `timedCalls` **only** when present — untimed calls do not inflate averages
- `_daily` bucket keyed by today (`YYYY-MM-DD`), aggregates across categories
- Guard: categories prefixed with `test-` or `_test` are skipped (no metrics.json mutation)
- No-op for `null` / `undefined` usage — file is not created
- `metrics.json` is created on first call when missing (via `mutateJsonFileLocked` default value)
- Unrelated `_engine` and `_daily` entries are preserved on every write
- 25 sequential calls yield exact totals (atomic via `mutateJsonFileLocked`)
- Swallows write errors without throwing (logs to `console.error`) — verified by replacing `metrics.json` with a directory

### `isResumeSessionStillValid(result)`

Additional edge cases beyond the existing happy paths:
- `undefined` result returns `false`
- Empty-string `sessionId` falls through to `raw` scan (and a marker in `raw` still keeps the session alive)
- `raw` without `\"session_id\"` marker returns `false`
- Non-string `raw` values (`false`, `0`, `null`) handled gracefully — guard uses truthy check before `.includes`
- Truthy `sessionId` short-circuits — a missing marker in `raw` does not flip the answer

### Module export shape

- `callLLMStreaming` is exported (was untested)
- Module surface matches exactly: `{ callLLM, callLLMStreaming, isResumeSessionStillValid, trackEngineUsage }`

## Test isolation

All `trackEngineUsage` tests use `createTestMinionsDir()` + re-require pattern to redirect writes to a temp `MINIONS_TEST_DIR`. The real `engine/metrics.json` is never mutated. Verified post-run that `metrics.json` contains no test category keys and no 1999-01-01 `_daily` seed.

Hermetic: no `claude` CLI spawn, no network I/O, no external deps.

## Test plan

- [x] \`npm test\` passes: **2523 passed, 0 failed, 3 skipped** (up from 2504)
- [x] Verified real \`engine/metrics.json\` unchanged after run
- [x] New tests run in < 1s (no process spawn)

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -3160,6 +3160,284 @@ async function testLlmModule() {
     assert.ok(src.includes('maxLineBufferBytes'), 'Should cap the incremental line buffer');
     assert.ok(src.includes('const toolUses = []'), 'Should retain structured tool-use metadata');
   });
+
+  // ── Export Shape ─────────────────────────────────────────────────────────
+  await test('llm module exports callLLMStreaming', () => {
+    assert.ok(typeof llm.callLLMStreaming === 'function');
+  });
+
+  await test('llm module exports exactly the documented surface', () => {
+    const exported = Object.keys(llm).sort();
+    assert.deepStrictEqual(exported, [
+      'callLLM',
+      'callLLMStreaming',
+      'isResumeSessionStillValid',
+      'trackEngineUsage',
+    ]);
+  });
+
+  // ── trackEngineUsage — persistence, accumulation, guards ─────────────────
+  // These tests run against a temp MINIONS_TEST_DIR so they do not pollute
+  // real engine/metrics.json. The module is re-required after createTestMinionsDir
+  // so shared.MINIONS_DIR / llm.ENGINE_DIR resolve against the temp tree.
+
+  await test('trackEngineUsage initializes a new category with all fields', () => {
+    const restore = createTestMinionsDir();
+    try {
+      const freshLlm = require(path.join(MINIONS_DIR, 'engine', 'llm'));
+      const metricsPath = path.join(process.env.MINIONS_TEST_DIR, 'engine', 'metrics.json');
+      freshLlm.trackEngineUsage('agent-dispatch', {
+        costUsd: 0.05, inputTokens: 1000, outputTokens: 500,
+        cacheRead: 200, cacheCreation: 100, durationMs: 3000,
+      });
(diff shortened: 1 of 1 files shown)


yemi33 (commented):
VERDICT: APPROVE

Solid TDD coverage for `engine/llm.js` — tests exercise actual behavior (not source-string matching) and align with the module at HEAD.

**Strengths**
- Export shape test locks the public surface (prevents silent additions).
- `trackEngineUsage` covered end-to-end: new-category init, accumulation, missing fields default to 0, durationMs/timedCalls only incremented when provided (llm.js:29-32), `_daily` bucket creation + cross-category aggregation, `test-`/`_test` skip guard (llm.js:15), null/undefined no-op, metrics.json auto-creation, unrelated category preservation, and a 25-call sequential stress test for atomic writes.
- Error-swallow test forces a write failure by replacing metrics.json with a directory — clean way to hit the `catch` at llm.js:47 without monkey-patching.
- `isResumeSessionStillValid` edge cases: `undefined` result, empty-string sessionId falling through to raw scan, missing marker, non-string raw (false/0/null — guarded by truthy check at llm.js:272-273), and sessionId-preferred-over-raw. Matches the function exactly.
- Uses `createTestMinionsDir()` + fresh `require()` so `ENGINE_DIR` resolves against the temp tree — no pollution of real `engine/metrics.json`.

**Minor nits (non-blocking)**
- The stress test label says "atomic writes" but is sequential; it validates no dropped writes through the lock path, which is the right property — wording could be "sequential accumulation via file lock."
- `fs.rmdirSync(metricsPath)` cleanup after the error-swallow test is defensive but technically unnecessary since `restore()` nukes the whole temp tree. Harmless.

No security/injection concerns — no user input, pure fs + JSON round-trip in a temp dir.

PR #1739, opened 2026-04-21
test(consolidation): add unit tests for inbox consolidation helpers

yemi33:
## Summary
Adds coverage for previously untested functions in `engine/consolidation.js`.

- `checkDuplicateHash`: undefined input, single-item, 100% duplicates, items missing `content` field
- `classifyToKnowledgeBase`: KB directory creation, category routing for all five KB categories, null/undefined content, agent=unknown fallback, `kb-checkpoint.json` write, unique filenames on title collisions, verbatim body preservation
- `consolidateInbox`: empty inbox no-op, sub-threshold no-op, `inboxConsolidateThreshold` config override, pinned-note exclusion, duplicate-hash circuit-breaker archive path, non-markdown file filtering, edge-case filenames without agent prefix

All behavioral tests use `createTestMinionsDir()` for isolation and avoid the LLM spawn by staying under the threshold or routing through the content-hash circuit breaker.

Part of daily coverage rotation — target: `engine/consolidation.js`.

## Test plan
- [x] `node test/unit.test.js` → 2526 passed, 0 failed, 3 skipped locally
- [x] No new external dependencies; uses the existing custom async runner
- [x] Each new test is atomic and resets module state via `createTestMinionsDir()`

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -2283,6 +2283,373 @@ async function testContentHashCircuitBreaker() {
     const result = checkDuplicateHash(items);
     assert.strictEqual(result.isDuplicate, true);
   });
+
+  await test('checkDuplicateHash returns false for undefined input', () => {
+    assert.strictEqual(checkDuplicateHash(undefined).isDuplicate, false);
+  });
+
+  await test('checkDuplicateHash returns false for a single item', () => {
+    // One item is never >80% of itself in the circuit-breaker sense (1/1 = 100%
+    // but the check is on >0.8 ratio; however a single item hash comparison is
+    // meaningless — document intended behavior: single item still reports
+    // isDuplicate=true because 1/1 > 0.8, but count/total encode reality.
+    // This test locks in the *current* semantics so regressions surface.
+    const result = checkDuplicateHash([{ name: 'a.md', content: 'only one' }]);
+    assert.strictEqual(result.isDuplicate, true);
+    assert.strictEqual(result.count, 1);
+    assert.strictEqual(result.total, 1);
+  });
+
+  await test('checkDuplicateHash reports correct count/total for 100% duplicates', () => {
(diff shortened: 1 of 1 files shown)


yemi33 (commented):
VERDICT: APPROVE

Solid coverage for consolidation.js helpers. The new tests exercise `checkDuplicateHash`, `classifyToKnowledgeBase`, and `consolidateInbox` behavior through real filesystem side-effects (via `createTestMinionsDir`) rather than string-matching the source — matches the behavioral-test convention from notes.md.

Strengths:
- Uses existing isolation helper (`createTestMinionsDir` + restore in `finally`) consistently, preventing leaks into real `prd/`/`knowledge/`.
- Covers both happy paths (category routing, frontmatter, checkpoint) and edge cases (null/undefined content, unparseable agent prefix, title-slug collisions → uniquePath, non-md files filtered).
- `consolidateInbox` tests hit the key branches without needing to stub the LLM: sub-threshold no-op, threshold override, pinned-note exclusion, and the duplicate-content circuit-breaker archive shortcut. Smart.
- Single-item `checkDuplicateHash` test explicitly documents the current (quirky) semantics as regression lock-in — good call-out in the comment.

Non-blocking observations:
- `test/unit.test.js:2288` — the single-item comment is slightly misleading (`1/1 = 100%` so it IS `> 0.8`); the final assertion matches reality, just the prose is a little self-contradicting. Cosmetic.
- The duplicate-archive test relies on the 5-file default threshold — fine today, but if `ENGINE_DEFAULTS.inboxConsolidateThreshold` ever changes, this test will silently pass with a different code path. Not worth blocking.

Test suite: 2525 pass, 1 fail. The sole failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`) is unrelated to this diff and pre-exists on master.

PR #1743, opened 2026-04-22
test(ado): add unit tests for pure helpers

yemi33:
## Summary

Adds coverage for previously untested pure helper functions in `engine/ado.js`.

Part of daily coverage rotation — target: `engine/ado.js`.

**New tests (36 total):**

- `classifyBuildStatus` (10): empty, all-passed, `partiallySucceeded` (warnings count as pass), single-failed, `canceled` == failed, `inProgress`/`notStarted` → running, mixed passed+inProgress → running, failed+inProgress → running (not failing — only terminal once `allDone`), unknown-result → none.
- `votesToReviewStatus` (11): empty → pending; +10/+5 → approved; -10 → changes-requested; -5 → waiting; all-zero → pending; **rejection (-10) overrides approval (+10)** per ADO semantics; approval beats waiting; sub-threshold (+4) → pending; waiting resolves only when no approval.
- `isAdoAuthError` (7): 401/403/HTML-redirect → true; 500/network/null/undefined/empty-message/unrelated → false; adjacent HTTP codes (400, 404) correctly return false.
- `needsAdoPollRetry` (3): initial state false, idempotent, independent of throttle state.
- Throttle state machine (4): `_resetAdoThrottle` returns to initial state; `_setAdoThrottleForTest` round-trips through `getAdoThrottleState`; partial overrides preserve un-named fields (Object.assign semantics); reset is idempotent.

## Implementation note

`classifyBuildStatus` and `votesToReviewStatus` are **not** exported from `engine/ado.js`. Rather than widening the public API for testing only, the tests extract the function source via regex and reconstruct them with `new Function(...)` — safe because both are pure and close over nothing in the module scope. Pattern documented in the inbox note.

## Test plan

- [x] `node test/unit.test.js` — **2539 passed**, 1 pre-existing failure (`Metrics JSON has valid structure: dallas missing tasksCompleted` — unrelated, verified by stashing diff and reproducing on master), 3 skipped.
- [x] New tests isolated via fresh `require.cache` bust at both ends of the `testAdoPureHelpers` block — no state leak into neighbouring test functions.
- [x] Every test that mutates throttle state calls `_resetAdoThrottle()` in teardown, matching the existing pattern (`test/unit.test.js:21681`, `:21715`, etc.).

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -11930,6 +11930,10 @@ async function main() {
     // W-mnyao4dyz8w7: createThrottleTracker factory, adoFetchText throttle, GitHub throttle
     await testCreateThrottleTracker();
     await testAdoFetchTextThrottle();
+
+    // W-moa4gs7dpd7a: ado.js pure helpers — classifyBuildStatus, votesToReviewStatus, isAdoAuthError, throttle
+    await testAdoPureHelpers();
+
     await testGhThrottle();
     await testGhThrottleEngineGuards();
     await testGhThrottleDashboard();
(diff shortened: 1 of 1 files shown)


yemi33 (commented):
VERDICT: APPROVE

Clean additive test PR — 275 lines, all in test/unit.test.js. No production code touched.

**Coverage verified:**
- classifyBuildStatus: empty, all-succeeded, partiallySucceeded, mixed-fail, canceled, inProgress, notStarted, mixed-running, and the subtle 'failed+inProgress → running' case (doesn't short-circuit until allDone).
- votesToReviewStatus: empty, +10, +5, -10, -5, all-zero, rejection-overrides-approval, approval-beats-waiting, sub-threshold (+4).
- isAdoAuthError: 401/403/HTML-redirect true-paths; 400/404/500/null/undefined/{}/unrelated false-paths (guards the unanchored regex against adjacent status codes).
- needsAdoPollRetry: initial false + idempotent + independent of throttle state.
- Throttle test helpers: reset, set/get roundtrip, partial override preservation, idempotent reset.

**Nice touches:**
- Fresh require + cache delete on entry and exit keeps module-level state (_adoPollHadAuthFailure, _adoThrottle) from leaking between tests.
- classifyBuildStatus/votesToReviewStatus aren't exported, so they're extracted via regex + `new Function` — pragmatic and keeps the real source path under test rather than mirroring logic. Functions are pure with no closure deps, so this is safe.
- Final teardown block resets throttle and busts require cache.

**Verified:** Ran full suite on the branch — 2539 passed, 1 unrelated pre-existing failure (`Metrics JSON has valid structure: dallas missing tasksCompleted`), all new W-moa4gs7dpd7a tests green.

No blocking issues.

PR #1748, opened 2026-04-23
test(timeout): add unit tests for idle/steering/timeout paths

yemi33:
## Summary

Adds **30 behavioral unit tests** for `engine/timeout.js` covering the three functions that previously had near-zero behavioral coverage (existing tests were source-code grep assertions):

- **`checkIdleThreshold`** (8 tests) — agent-idle detection, pending-work short-circuit, threshold logic, `_idleAlertSent` dedupe, recovery reset
- **`checkSteering`** (8 tests) — session-present kill path, no-session inbox forward path, empty-file handling, double-kill guard, 30s retry with `_steeringRetried` idempotency
- **`checkTimeouts`** (14 tests) — hard timeout, per-item `meta.deadline` override, orphan detection, `engineRestartGraceUntil`/`engineRestartGraceExempt`, hung tracked process, steering skip window, reconcile retry/fail/preserve/done-guard, completion-via-output scan (success + error subtypes)

## Design

Uses **`require.cache` injection** to install a fake engine module so tests drive the full function body without loading the real `engine.js` (which has heavy transitive requires and wouldn't cache-bust cleanly):

```js
require.cache[enginePath] = {
  id: enginePath, filename: enginePath, loaded: true,
  exports: { activeProcesses: new Map(), realActivityMap: new Map(),
             engineRestartGraceUntil: 0, engineRestartGraceExempt: new Set() },
  children: [], paths: [],
};
```

`lifecycle` is stubbed the same way. `shared.killImmediate` / `killGracefully` / `exec` are replaced on the fresh shared instance with no-op counters so tests never spawn real kill commands.

## Test plan

- [x] `npm test` — 2533 passed (was 2503), +30 new, 0 new failures
- [x] All 30 new tests pass deterministically in full-suite run
- [x] No leak into unrelated tests — `require.cache` is restored in `finally` blocks
- [x] No real process kills — stubbed `shared.killImmediate`/`killGracefully`/`exec`

Pre-existing (unrelated) failure: `Metrics JSON has valid structure: dallas missing tasksCompleted` — the `dallas` entry in `engine/metrics.json` has only `prsApproved`, no `tasksCompleted`/`tasksErrored`. Fails both before and after this PR; flagged for separate cleanup.

Part of daily coverage rotation — target: `engine/timeout.js`

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -9340,6 +9340,752 @@ async function testAgentSteering() {
   });
 }
 
+// ─── engine/timeout.js — Behavioral Coverage ────────────────────────────────
+// Exercises checkIdleThreshold, checkSteering, checkTimeouts end-to-end using a
+// fake engine module injected into require.cache. Uses createTestMinionsDir for
+// filesystem isolation and stubs shared.killImmediate/killGracefully to keep
+// tests from spawning real kill commands.
+
+async function testTimeoutBehavioral() {
+  console.log('\n── engine/timeout.js — Behavioral ──');
+
+  const enginePath = require.resolve(path.join(MINIONS_DIR, 'engine'));
+  const timeoutPath = require.resolve(path.join(MINIONS_DIR, 'engine', 'timeout'));
+  const lifecyclePath = require.resolve(path.join(MINIONS_DIR, 'engine', 'lifecycle'));
+  const routingPath = require.resolve(path.join(MINIONS_DIR, 'engine', 'routing'));
+  const sharedPath = require.resolve(path.join(MINIONS_DIR, 'engine', 'shared'));
+  const queriesPath = require.resolve(path.join(MINIONS_DIR, 'engine', 'queries'));
+  const dispatchPath = require.resolve(path.join(MINIONS_DIR, 'engine', 'dispatch'));
+
+  // Build an isolated test environment: fresh MINIONS dir, fresh shared/queries/
+  // dispatch/timeout/routing modules, fake engine + stubbed lifecycle injected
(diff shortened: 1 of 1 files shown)


yemi33 (commented):
VERDICT: APPROVE

Solid behavioral coverage for engine/timeout.js (+748 lines, test-only).

**Strengths:**
- Uses the documented require.cache injection pattern for the engine↔timeout circular dep — fake engine + stubbed lifecycle keep the tests hermetic.
- Filesystem isolation via `createTestMinionsDir` and proper `env.restore()` in `finally` for each test — no bleed between tests.
- Stubs `killImmediate`/`killGracefully`/`exec` with counters, so no real processes are spawned.
- Exercises actual code paths (not source-string matching) — matches the team convention from notes.md (behavioral tests > source inspection).
- Good branch coverage: idle threshold (busy/pending/idle/sent-guard/recovery/custom-threshold/empty-agents/default), steering (no-session, session, empty msg, double-kill guard, stale-retry, already-retried), timeouts (hard, meta.deadline, orphan, grace, grace-exempt, hung tracked, recently-steered skip, reconcile retry, reconcile exhausted, reconcile live, reconcile done, completion-via-output success/error).
- Uses constants (`shared.ENGINE_DEFAULTS.maxRetries`) instead of hardcoded 3.

**Minor non-blocking observations:**
- The three 120ms sleep-based idle tests add ~360ms to the suite — acceptable given the value, but if suite runtime becomes a concern, consider exposing `_lastActivityTime` as injectable.
- `assert.ok(/Orphaned|silent/.test(completed.reason || ''))` is lenient by design; fine.

Nothing blocking. Good shape.

PR #1323, opened 2026-04-18
feat(SEC-03 Phase A): escapeHtml helper + hotspot XSS migration + regression gate

yemi33:
## Summary

Implements PRD item **P-d4a8f6c2** — SEC-03 Phase A — per meeting decision D-3 (*helper + top-10 + migrate-on-touch*). Phase A lands the foundation and migrates the six hottest dashboard renderers so an attacker-controlled work-item title / PR title / KB entry cannot script the dashboard. Phase B (remaining ~20 files) is deferred to a follow-up.

### What shipped

1. **`escapeHtml(str)`** in `dashboard/js/utils.js` — 6-character escape (`& < > " ' /`), null/undefined → `''`, idempotent for non-metacharacter input. The `/` escape closes the `</script>` break-out that the legacy 5-char `escHtml` left open.
2. **Hotspot renderers migrated** to `escapeHtml` for every user-controlled field reaching `.innerHTML`:
   - `render-work-items.js` (title, description, agent, PR, refs, criteria)
   - `render-agents.js` (card fields; detail header rebuilt via `replaceChildren` + `textContent`)
   - `render-prs.js` (title, agent, branch, reviewer)
   - `render-kb.js` (title, agent, preview, category)
   - `render-plans.js` (summary, project, generator, E2E PR links, version labels)
   - `render-inbox.js` (name, age, preview, pin key)
3. **Regression gate** in `test/unit.test.js` counts dynamic `.innerHTML =` / `.innerHTML +=` assignments across `dashboard/` and asserts the total ≤ **172** (frozen Phase A baseline). Static literal RHS exempt; any new unsafe assignment fails CI; `textContent` conversions can lower the baseline.

`escHtml` is kept as a named backward-compat shim so unmigrated Phase B files still compile — it will be removed once Phase B lands.

## Acceptance criteria

- [x] `escapeHtml` exported from `dashboard/js/utils.js` with unit tests for all 6 metacharacters, null/undefined, and idempotency
- [x] Regression test asserts dynamic innerHTML count ≤ baseline; passes on current tree
- [x] All listed hotspot renderers escape user-controlled fields before innerHTML interpolation
- [x] Baseline reflects post-migration count (172) not pre-migration
- [x] Manually verified: `` renders as literal text

## Test plan

- [x] `npm test` — 2460 pass / 1 fail (pre-existing, unrelated `dallas missing tasksCompleted` from live metrics.json; confirmed failing on master before this PR)
[...]

PR #1751, opened 2026-04-23
fix: escape </script> literal in utils.js (closes #1746)

### LINKED ISSUE #1746: Unescaped </script> block in utils.js breaking dashboard
The </script> tag in dashboard/js/utils.js on line 101 is not escaped, causing leaking js in the dashboard

yemi33:
Closes #1746.

## The bug

`dashboard/js/utils.js:101` contained the token `</script>` verbatim inside a line comment (introduced by the SEC-03 Phase A escapeHtml work, PR #1323). `dashboard.js` inlines every `dashboard/js/*.js` module into a single `<script>/* __JS__ */</script>` block in `layout.html`. The HTML5 tokenizer's script-data end-tag scan is byte-level and does **not** honour JS comment or string boundaries — so that literal inside the comment terminated the inline `<script>` block early, and the remainder of the assembled dashboard JS spilled into the document as visible text, breaking the whole dashboard.

## The fix

Rewrite the offending comment (and the new NOTE added beside it) so the token is spelled `<\/script>` instead of `</script>`. After `<` the HTML parser enters *script-data-less-than-sign state*, which only recognises `/` (end-tag-open) or `!` (escape-start). `\` is "anything else" — it emits `<` back to script-data state, so the end-tag match never starts. The backslash is meaningless inside a JS line comment (just bytes), so there is no JS-side effect.

The original `escapeHtml` implementation itself was already correct — only the documentation comment was unsafe.

## Regression gate

Added a unit test in `test/unit.test.js` (inside `testSec03EscapeHtml`) that walks `dashboard/js/` and fails if any file contains the byte sequence `</script` followed by one of `\t \n \f \r ' ' '/' '>'` — the full HTML5 end-tag-close alphabet, not just `</script>` with a trailing `>`. This catches the same bug class in any future comment, string, or template literal, not just this one line.

**TDD confirmation:** the test fails on the pre-fix source (offender: `utils.js:101`) and passes after the fix.

## Test plan

- [x] Reproduced the bug by scanning `dashboard/js/*.js` — the pre-fix source shows `offenders: [ 'utils.js:101' ]`.
- [x] Applied the fix, re-ran the same scan — `offenders: []`.
- [x] Simulated the full inline-script concatenation from `dashboard.js`'s `jsFiles` list and confirmed zero premature `</script` tokens in the assembled body.
- [x] `npm test` — **2799 passed / 0 failed / 3 skipped**, including the new regression test and all existing SEC-03 escapeHtml tests.

### DIFF
--- a/dashboard/js/utils.js
+++ b/dashboard/js/utils.js
@@ -98,9 +98,15 @@ function toggleModalPin() {
 
 // Canonical HTML-escape helper (SEC-03). Use this in all new code and for any user-controlled
 // field that reaches `.innerHTML` / a template literal. Escapes the 6 HTML metacharacters
-// (& < > " ' /) — the `/` escape closes the `</script>` break-out path that a 5-char escape
+// (& < > " ' /) — the `/` escape closes the `<\/script>` break-out path that a 5-char escape
 // leaves open. Returns '' for null/undefined so missing fields never render the literal strings
 // "null"/"undefined". Idempotent for non-metacharacter input (double-escaping only expands `&`).
+// NOTE: the `<\/script>` spelling above is deliberate. dashboard.js inlines every module in
+// dashboard/js/ into a single inline <script> block; a raw closing-script-tag literal in any
+// comment or string closes that block early and spills the rest as document text (issue #1746).
+// The HTML5 tokenizer's script-data end-tag match is byte-level and ignores JS comment/string
+// boundaries, so the only safe way to reference the token in-source is to break the match —
+// `<` followed by `\/` works because after `<` only `/` (not `\`) triggers end-tag-open state.
 function escapeHtml(str) {
   if (str === null || str === undefined) return '';
   return String(str)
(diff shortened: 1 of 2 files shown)


yemi33 (commented):
VERDICT: APPROVE

Surgical, well-diagnosed fix.

**What's right:**
- Root cause nailed: HTML5 script-data end-tag scan is byte-level and ignores JS comment/string boundaries, so the literal `</script>` in a comment at `dashboard/js/utils.js:101` closed the inlined `<script>` block early. Fix (`<\/script>`) uses the `script-data-less-than-sign` → anything-else transition correctly — backslash emits `<` back and kills the end-tag match without any JS-side effect (it's just bytes inside a `//` comment).
- Regression test is better than the bug: regex `/<\/script[\t\n\f\r \/>]/` matches the **full** HTML5 end-tag-open alphabet, not just `</script>` with a trailing `>`. Catches `</script `, `</script/`, etc.
- Test's own explanatory comment uses `<\/script>` — so the regression gate doesn't trip on itself. Verified.
- No behavioral change to `escapeHtml`; only comment/docs text changed plus a new NOTE explaining why the spelling is deliberate (preventing re-introduction).
- Real diff is 45 lines across 2 files — matches PR description. Earlier stat confusion was my stale local `master`; `origin/master...` confirms the tight scope.

**Minor non-blocking:**
- The test's `endTagRe.test(src)` → `lines.findIndex(l => endTagRe.test(l))` re-scans; fine given tiny file set, not worth changing.
- `escHtml` legacy alias retained — comment flags Phase B cleanup, matches the deprecation convention.

Ship it.

